PR #124, opened 2020-05-16
Enable CodeQL scanning

MKusber:
This PR will enable a github action triggering codeQL checks in this repository

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ If you want to use Docker-based deployment, you need to install Docker on your l
 
 #### Running the Full cwa Backend Using Docker Compose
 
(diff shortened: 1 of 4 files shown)


jhutchings1:
👋🏻  Hey folks, glad to see you're using code scanning! A couple of comments I wanted to share: 

- We include the cron trigger to ensure that there's always at least one analysis per week. While that's redundant while there's active development going on, it may be the only analysis which happens if active development ceases. Since the threat landscape is constantly changing, this ensures you learn about new threats in either case. 
- For the push event, we recommend that you run on all branches so that you can learn about newly introduced security issues in your pull requests. Running only on master means that the vulnerabilities can be introduced to your mainline before we detect them. 

In any case, thanks for testing this out and please feel free to shout if you have any concerns!

PR #174, opened 2020-05-18
Update Copyright headers

michael-burwig:


AndreasGB:
Does that mean that Deutsche Telekom AG was removed / removed itself from the project or are they still a part of it?

johanneseschrig:
Hi @AndreasGB! No - this change does not imply that. Since the code in this repository is maintained by SAP developers, we decided to simplify the header files here.

AndreasGB:
Makes sense, thank you!

PR #177, opened 2020-05-18
Bumps version to 0.3.3

johanneseschrig:


pithumke:
... and in the dev/int deployments after merging.

johanneseschrig:
@ole-lilienthal I recommend the [files](https://github.com/corona-warn-app/cwa-server/pull/177/files) tab ;)

johanneseschrig:
@pithumke DEV yes, INT will not be affected yet, since it points to `integration`.

ole-lilienthal:
I see, I see, my bad. Infra should be fine without adjustments.

Issue #168, opened 2020-05-18
Implement final Google/Apple server API and file specifications


pithumke:
## Current Implementation
Distribution and submission services implement our own API and file definitions.

## Suggested Enhancement
Implement final Google/Apple server API and file specifications, once a final version is available.

## Expected Benefits
- Being compliant to the "official" Google/Apple specs
- Potentially enabling "pan-european scenario"
- Remove file conversion from apps (by already outputting the files in the format expected by the Google/Apple client APIs).

christian-kirschnick:
https://static.googleusercontent.com/media/www.google.com/en//covid19/exposurenotifications/pdfs/Exposure-Key-File-Format-and-Verification.pdf

mh-:
If you consider replicating the Google reference implementation ("Google has created an [implementation of an EN server](https://github.com/google/exposure-notifications-server) that you can use as a reference"), which uses "Device Attestation", please also consider potential privacy impact of Device Attestation - see https://github.com/corona-warn-app/cwa-documentation/issues/62

christian-kirschnick:
Hey @mh-,

although we fill try to stick to the proto definitions from Google/Apple as close as possible for the backend implementation, it does not necessarily mean we will process/send all fields in the protocol. There are a number of fields which are heavily debated within DPP (device attestation just being one of them), and DPP drives a lot of the implementation decisions in this project.

The backend topic for device attestation is tracked in https://github.com/corona-warn-app/cwa-server/issues/82.

cheers
Chris

PR #182, opened 2020-05-19
Reduce size of exposure notifications

Spacefish:
Well, we probably can´t determine an infection time down to the millisecond, so seconds should be fine and we can jump to an int32 instead of an int64.

This saves 8 Byte per exposed user when downloading the exposure notifications.. Especially because the milliseconds part contains a lot of entropy and can´t be compressed efficiently.

Assuming 5k exposed users, this saves 39KByte/download when we assume ~60 Million users this is a lot of traffic!

Hopefully Corona won´t be an issue in 2038 ;)

### DIFF
--- a/common/protocols/src/main/proto/app/coronawarn/server/common/protocols/external/exposurenotification/exposure_notification.proto
+++ b/common/protocols/src/main/proto/app/coronawarn/server/common/protocols/external/exposurenotification/exposure_notification.proto
@@ -11,8 +11,8 @@ message File {
 }
 
 message Header {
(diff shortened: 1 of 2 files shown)


christian-kirschnick:
Hey @Spacefish,

thanks a lot for your first PR! Unfortunately, we won't be able to merge your changes, since we are currently in the process of adapting to the newest protocol specification published by Apple/Google. CWA will follow the specification of Apple/Google as close as possible in order to ensure support of future federation scenarios as well.

You can find more information here :https://github.com/corona-warn-app/cwa-server/issues/168. We will be having a PR for this issue soon.

cheers & thanks again,
Chris

PR #194, opened 2020-05-20
Use maven wrapper as the single source of truth for maven version in the project

driptaroop:
#### Currently this project does not use any standard maven version. Different kind of build is done by different maven installation (which might or might not be of same version). For example,

1. maven wrapper
2. CI platforms (`circleci` and `github actions`) use the maven version provided by the respective CI platform
3. `Dockerfile` uses `maven` official docker image with version  `3.6.3`

# This might lead to issues such as:

1. Assumption that the CI supplied maven version is same one that the project was build on in local
2. Different developers using different maven version
3. To update the maven version in the future, the developers have to remember to update it across all the different places mentioned above, failing / forgetting to do all of them might lead to build discrepancy and unreliable builds.
4. Edge cases might occur where build may fail because of maven version mismatch.

# What this PR does:

1. Standardises this project on a given Maven version (as per wrapper), leading to more reliable and robust builds.
2. Provisioning a new Maven version to different users and execution environment (e.g. IDEs or CI) is as simple as changing the Wrapper definition.
3. Since the wrapper was already included in the project I just changed the different usages of maven to point to the maven wrapper instead.


Bananeweizen:
Unifying the maven versions via wrapper is all good, but it does not prevent developers using another maven version by invoking maven without the wrapper. I suggest to additionally use the maven enforcer in the project to require a maven runtime version: https://maven.apache.org/enforcer/enforcer-rules/requireMavenVersion.html

psavva:
I am not able to build with docker-compose with this revision: d7debae1cf2b6ed5d821a71ee6b6b36a01d9cd63

please run docker-compose up from the command line.
It fails with mvnw command.

Please note that i'm running this from a windows host to build

stevesap:
> I am not able to build with docker-compose with this revision: [d7debae](https://github.com/corona-warn-app/cwa-server/commit/d7debae1cf2b6ed5d821a71ee6b6b36a01d9cd63)
> 
> please run docker-compose up from the command line.
> It fails with mvnw command.
> 
> Please note that i'm running this from a windows host to build

Please run `git checkout-index --force --all` This should force the [changes to the `.gitattributes`](https://github.com/corona-warn-app/cwa-server/commit/d7debae1cf2b6ed5d821a71ee6b6b36a01d9cd63#diff-fc723d30b02a4cca7a534518111c1a66) to take effect which are necessary on windows machines due to EOL fix (otherwise, a fresh clone would also work)

Issue #196, opened 2020-05-20
[BSI][0.3.3] Missing Check of Return Value


BSI-TF-CWA:
Rating: Low
 
Description:
A missing check of a return value could lead to unexpected behavior. The method createNewFile() returns a boolean (see https://docs.oracle.com/javase/7/docs/api/java/io/File.html#createNewFile()) which is not checked. The function would return false if the file already exists. In the current implementation this is not an issue (file name is a constant) but could lead to unexpected behavior in the future. It could potentially lead to unintentionally overwriting existing files. The function name and the function doku should indicate that this function ignores existing files.
 
Proof of Concept:
The offending line can be found under: https://github.com/corona-warn-app/cwa-server/blob/82d2ff55d604f122c7b003adce358979bc7bc7a2/services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/io/IO.java#L43-L48

michael-burwig:
Hi @BSI-TF-CWA,
thank you for bringing this to our attention.

We identified throwing an appropriate Exception as the desired behavior if the file in question already exists. Hence, processing should terminate at this point.

pithumke:
Dear @BSI-TF-CWA,
the issue was addressed through community PR #197. Big thank you again to contributor @Alex1607!

PR #197, opened 2020-05-20
Addressed issue #196

Alex1607:
* Added a return Value
* Renamed methode
* Adjusted documentation


services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/io/IO.java
```diff
@@ -33,15 +33,16 @@
...
    * @param root The parent file.
    * @param name The name of the new file.
+   * @return Whether the file is successfully created
    */
-  public static void makeFile(File root, String name) {
+  public static boolean makeNewFile(File root, String name) {
     File directory = new File(root, name);
     try {
-      directory.createNewFile();
+      return directory.createNewFile();
```
michael-burwig, line 45:
```suggestion
      if (!directory.createNewFile()) {
        throw new IOException("Could not create " + name + ", file already exists");
      }
```

PR #201, opened 2020-05-20
Validate Diagnosis Keys before they are distributed

johanneseschrig:
- Refactors validation to use annotations & Validator
- Adds check to retrieved keys to prevent distributing tampered keys


common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/DiagnosisKey.java
```diff
@@ -39,13 +51,26 @@
...
 
+  private static final Logger logger = LoggerFactory.getLogger(DiagnosisKey.class);
+
+  private static final Validator VALIDATOR = Validation.buildDefaultValidatorFactory().getValidator();
+
   @Id
   @GeneratedValue(strategy = GenerationType.IDENTITY)
   private Long id;
+
+  @Size(min = 16, max = 16, message = "Key data must be byte array of length 16")
```
MKusber, line 62: Please add a period to the message

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/DiagnosisKey.java
```diff
@@ -128,6 +153,35 @@ public boolean isYoungerThanRetentionThreshold(long daysToRetain) {
...
+    return violations.isEmpty();
+  }
+
+  /**
+   * Validates this key.
+   */
+  public void validate() {
+    Set<ConstraintViolation<DiagnosisKey>> violations = getConstraintViolations();
+
+    if (!violations.isEmpty()) {
```
MKusber, line 172: Shouldn't we reuse `isValid` here?

common/persistence/src/test/java/app/coronawarn/server/common/persistence/domain/DiagnosisKeyTest.java
```diff
@@ -59,6 +64,111 @@ public void testSubmissionTimestampGetter() {
...
+        "[Rolling period must be greater than 0. Invalid Value: -3]");
+  }
+
+  @Test
+  public void rollingPeriodDoesNotThrowForValid() {
+    Assertions.assertThatCode(keyWithRollingPeriod(144L)::validate).doesNotThrowAnyException();
+  }
+
+  @Test
+  public void keyDataMustHaveValidLength() {
```
michael-burwig, line 128: better parameterized

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/validation/ValidRollingStartNumberValidator.java
```diff
@@ -0,0 +1,33 @@
...
+import java.time.LocalDateTime;
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+
+public class ValidRollingStartNumberValidator implements ConstraintValidator<ValidRollingStartNumber, Long> {
+
+  @Override
+  public boolean isValid(Long rollingStartNumber, ConstraintValidatorContext constraintValidatorContext) {
+    long currentInstant = LocalDateTime.ofInstant(Instant.now(), UTC).toEpochSecond(UTC) / 600L;
+    return rollingStartNumber > 0L && rollingStartNumber < currentInstant;
```
MKusber, line 31: This still seems too broad to me. I have issues accepting keys from the 80s should be valid.

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/DiagnosisKey.java
johanneseschrig: I guess we want to log out the violations, hence we would need to call the `getConstraintViolations` again if we called `isValid` here.

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/validation/ValidRollingStartNumberValidator.java
johanneseschrig: [We don't](https://github.com/corona-warn-app/cwa-server/blob/feature/key-validation-distribution/services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java#L160): A key from the 80s can still be semantically valid - it will just not arrive in our database.

MKusber: I missed this. Sorry

PR #211, opened 2020-05-20
Increase robustness of DiagnosisKeyService: reject negative retention period

jonaskoepper:
_Upfront: Great project and thank you for the contribution so far, colleagues!_

---
I could provide a small improvement to code robustness (here by applying the principle of _failing early_):

Service bean [`DiagnosisKeyService`](https://github.com/corona-warn-app/cwa-server/blob/master/common/persistence/src/main/java/app/coronawarn/server/common/persistence/service/DiagnosisKeyService.java#L35) exposes `public` method `applyRetentionPolicy(int daysToRetain)`

- As a public method, it should validate input parameters.
- Though, as of now, _negative_ values for the diagnosis key retention period are allowed.

The method concerned is the following:

https://github.com/corona-warn-app/cwa-server/blob/90f713db6e3f2207e167ad82a9b9857415fcf4d9/common/persistence/src/main/java/app/coronawarn/server/common/persistence/service/DiagnosisKeyService.java#L61-L73

- While a negative integer theoretically could be considered as valid input, it is questionable if there is the requirement to delete diagnosis key entries having a timestamp in the future (as such entries should not exist in the first place).
- An alternative would be to accept negative integers for `daysToRetain` and to interpret this as a request to have all diagnostic keys deleted, equivalent to providing `0` as input parameter. I would argue against this alternative because it could hide an actual problem with the retention period value provided.

### DIFF
--- a/common/persistence/src/main/java/app/coronawarn/server/common/persistence/service/DiagnosisKeyService.java
+++ b/common/persistence/src/main/java/app/coronawarn/server/common/persistence/service/DiagnosisKeyService.java
@@ -62,12 +62,17 @@ public class DiagnosisKeyService {
   }
 
   /**
-   * Deletes all diagnosis key entries which have a submission timestamp that is older than the specified number of
-   * days.
+   * Deletes all diagnosis key entries which have a submission timestamp that is older than the
+   * specified number of days.
    *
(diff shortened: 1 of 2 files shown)


jonaskoepper:
@michael-burwig Merge conflicts are now resolved.

PR #212, opened 2020-05-20
Adjust architecture-overview.md

christian-kirschnick:


ole-lilienthal:
@christian-kirschnick prominent link on the readme would have been a good addition...

PR #215, opened 2020-05-21
remove WORKDIR from Docker build stage

CCFenner:
As all commands in the build stage can be executed with absolute paths, using `WORKDIR` or `cd ${WORK_DIR}` are not necessary. 

This solves [HaDoLint](https://github.com/hadolint/hadolint) findings [DL3003](https://github.com/hadolint/hadolint/wiki/DL3003) and [SC2164](https://github.com/koalaman/shellcheck/wiki/SC2164).


stevesap:
@CCFenner could you address linting issues?

`/dev/stdin:8 SC2086 Double quote to prevent globbing and word splitting.`

CCFenner:
@stevesap ignored [SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086) findings as `mvn` execution will not work as expected if values is quoted.

PR #222, opened 2020-05-21
Prevent multi line exception log messages

MKusber:
This is necessary to better cater the needs for connected monitoring systems. For the dev profile, nothing changes to ensure better readability in development.


services/submission/src/main/resources/log4j2dev.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
johanneseschrig, line 1: make file names consistent (log4j2-dev.xml)

MKusber: Thank you. Good catch

PR #232, opened 2020-05-21
Replace MinIO with CloudServer

### LINKED ISSUE #163: Open MinIO read access
## Current Implementation
Receiving files from local MinIO requires access key and secret key.

## Suggested Enhancement
Add a security policy to the MinIO configuration to only protect write access, but have open read access.

## Expected Benefits
Mobile colleagues can use local setup for end-to-end testing.

### LINKED ISSUE #176: Need to append "/index" to all requests on local setup
## Describe the bug
On the local setup (docker-compose), one needs to append `/index` to all GET requests.

## Expected behaviour
API as defined in the OpenAPI document.

## Technical details
This is a workaround to a ["fundamental design flaw"](https://github.com/minio/minio/issues/7335) in MinIO, through which files can not have the same name as directories. This is not an issue on "proper" S3 compatible storages (e.g. AWS or OpenShift) and only affects local development - on open shift, the files are properly renamed. However, this is still a deviation from the documented API.

## Possible Fix
**Option A**
- Remove MinIO and the "upload" distribution runner from the local setup, instead serve the files generated by the "assembly" runner from the local file system using something like nginx ([config from mock server can be reused](https://github.com/corona-warn-app/cwa-server/blob/6e7853c37c2a7a30bda104feefb50b29131335a5/tools/mock-server/nginx.conf)). This would enable easy end-2end scenario testing (submission to distribution), but would also mean that not all of our code is run locally.

**Option B:**
- Look for an alternative to MinIO that supports files and directories having the same name

**Option C:**
- Remove MinIO and force users to provide some S3 compatible bucket themselves (e.g. private AWS S3, etc.)

**Option D:**
- Add the "/index" suffixes to the actual API definition

**Option E:**
- Just leave this as is for now, wait for Apple/Google to finally release a backend API spec and then implement that instead. Any "fixing" of the current API probably needs to be thrown away in a few days anyways when they force an API spec onto us.

**Option F:**
- ???

ole-lilienthal:
Fixes #176 
Fixes #163


services/distribution/src/test/resources/application.properties
```diff
@@ -14,9 +14,10 @@ spring.flyway.locations=classpath:db/migration/h2
 spring.jpa.hibernate.ddl-auto=validate
 
 # S3 object store configuration
-cwa.objectstore.accessKey=${CWA_OBJECTSTORE_ACCESSKEY:minioadmin}
-cwa.objectstore.secretKey=${CWA_OBJECTSTORE_SECRETKEY:minioadmin}
-cwa.objectstore.endpoint=${CWA_OBJECTSTORE_ENDPOINT:http\://localhost\:9000}
+cwa.objectstore.access-key=${CWA_OBJECTSTORE_ACCESSKEY:minioadmin}
```
christian-kirschnick, line 17: Needs to be changed as well I guess?

start_cloudserver.sh
```diff
@@ -0,0 +1,2 @@
+
+docker run --name s3server -p 8000:8000 -e REMOTE_MANAGEMENT_DISABLE=1 -e SCALITY_ACCESS_KEY_ID=accessKey1 -e SCALITY_SECRET_ACCESS_KEY=verySecretKey1 zenko/cloudserver
```
christian-kirschnick, line 2: Port 8000 looks weird, because our (default) config is assuming 8003 now

ole-lilienthal: Thanks for pointing at the scripts. Will delete them since they are not needed anylonger.

services/distribution/src/test/resources/application.properties
ole-lilienthal: Fixed

start_cloudserver.sh
ole-lilienthal: Fixed

PR #236, opened 2020-05-21
Don't fail circleci build if sonar fails

johanneseschrig:
`mvn verify sonar:sonar` fails for forked PR builds. For now, just see if we can add the don't flag so sonar won't work for community PRs for now, but unblocks them.


CCFenner:
Github native SonarCloud integration could be a solution here: https://github.com/apps/sonarcloud

Issue #207, opened 2020-05-20
Externalize strings and configuration properties


pithumke:
There are quite a few hardcoded strings in the CWA server that should be externalized and/or turned into configurable properties. Example [1](https://github.com/corona-warn-app/cwa-server/blob/update/file-and-proto-specs/services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/structure/directory/decorator/signing/AbstractSigningDecorator.java#L40), [2](https://github.com/corona-warn-app/cwa-server/blob/update/file-and-proto-specs/services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/diagnosiskeys/structure/directory/DiagnosisKeysHourDirectory.java#L78), [3](https://github.com/corona-warn-app/cwa-server/blob/update/file-and-proto-specs/services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/diagnosiskeys/structure/file/TemporaryExposureKeyExportFile.java#L22), and many more...

pithumke:
Linking #183 and #238.

pithumke:
Closed by #183

Issue #183, opened 2020-05-19
Use configuration classes instead of @Value annotations


TilmannF:
You're using `@Value` about a dozen times, which isn't considered best practice anymore.

To quote Thomas Uhrig (@tuhrig):
> What will happen when you use this technique through your application? Well, whenever you need some property, you will inject it. You will inject it in services, controllers and components and you will properly inject the same property in different classes. That’s easy and absolutely what the mechanism is about.
> 
> However, this means nothing else than scattering your configuration about your whole application. Every class can pick a couple of properties to use. You will have no idea or control which class is using which properties. You will end-up with doing a full text search on your project to find out where a single key is used. You will have a lot of fun if you want to rename one of those keys or when you need to set each and every property for each and every class in your unit tests.
> [...]
> If you do so [use a configuration class], you will have a single point of responsibility to load and get your configuration from. As any other service, you can easily change the implementation. Maybe you don’t want to load your properties from a properties file, but from a database or web service!
([source](https://tuhrig.de/why-using-springs-value-annotation-is-bad/))

While this issue wouldn't qualify as high priority right now, it'll probably help in the future.

MALPI:
Hey there,

I'd like to contribute and work on this issue?

BTW: Great that you uses the `good first issue` label.

jwedel:
@MALPI You can have a look at the spring boot docs regarding the @ConfigurationProperties annotation:
- https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-enabling

The way to go would be to keep the property keys the same and use a separate class that binds the common key prefix (e.g. `services.submission`)  and provides type safe config values. Then inject that into the class that currently uses `@Value`.

See here for an example: 
- https://www.baeldung.com/configuration-properties-in-spring-boot

indyaah:
@MALPI : Are you still planning to work on this? If you're not planning to I would be happy to take over.

MALPI:
Hey,

I didn't find the time yet. Feel free to take it.

Gesendet von ProtonMail mobile

-------- Original-Nachricht --------
An 21. Mai 2020, 03:52, Anuj Patel schrieb:

pithumke:
[...]

PR #238, opened 2020-05-21
 #183: Use configuration classes instead of `@value` annotations 

indyaah:


services/distribution/src/main/java/app/coronawarn/server/services/distribution/runner/TestDataGeneration.java
```diff
@@ -81,8 +78,12 @@
   private static final long TEN_MINUTES_INTERVAL_SECONDS = TimeUnit.MINUTES.toSeconds(10);
 
   @Autowired
-  public TestDataGeneration(DiagnosisKeyService diagnosisKeyService) {
+  public TestDataGeneration(DiagnosisKeyService diagnosisKeyService,
+      DistributionServiceConfig distributionServiceConfig) {
     this.diagnosisKeyService = diagnosisKeyService;
+    this.retentionDays = distributionServiceConfig.getRetentionDays();
```
indyaah, line 84: Wasnt sure if this would be an acceptable pattern but started with it thinking this would be least intrusive change to service/controller layer. 

Would love to see/explore better design.

(Applies to most changes)

services/submission/src/main/java/app/coronawarn/server/services/submission/config/SubmissionServiceConfig.java
```diff
@@ -0,0 +1,65 @@
...
+  public Integer getRetentionDays() {
+    return retentionDays;
+  }
+
+  public void setRetentionDays(Integer retentionDays) {
+    this.retentionDays = retentionDays;
+  }
+
+  public Integer getMaxNumberOfKeys() {
+    return payload.getMaxNumberOfKeys();
```
indyaah, line 44: One more design decision that needs critical review/thought.

Idea was to abstract away complexities (additional classes/getters) in configuration structure and only expose single contract outside.

Applies to `DistributionServiceConfig` as well.

pithumke:
Hi @indyaah,
your PR looks really great so far, thanks for your support! Could you please do us a huge favor and also include configuration properties for the signature information, once #240 is merged? I tried to already incorporate `@ConfigurationProperties` in that PR but didn't have any luck with it (and can't look into it anymore because of time constraints). I pushed my attempt to https://github.com/corona-warn-app/cwa-server/commit/7d752e8ba17e41b20deb5c9299c14fe8539c9649 and added a comment here: https://github.com/corona-warn-app/cwa-server/issues/183#issuecomment-632358552.

indyaah:
@pithumke : Definitely, will incorporate 240 changes once it's merged. Thanks for the references.

services/submission/src/main/java/app/coronawarn/server/services/submission/config/SubmissionServiceConfig.java
jwedel: When abstracting the payload away, you will lose the information that `getMaxNumberOfKeys` refers to the payload. Then you would need to rename this to `getMaxNumberOfPayloadKeys`. Not sure if this is worth it. But if there is some redundant more complex operations done on the properties across the code, then this could be moved to this class.

Maybe [@ConstructorBinding](https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-constructor-binding) should be used to prevent in incomplete object that could result in NPEs (e.g. in `return payload.getMaxNumberOfKeys();).

Also, `@ConfigurationProperties` can be annotated with `@Validated` which will actually crash the app at start-up time with a proper readable error message in case non-optional properties are missing:

https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-validation

services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java
```diff
@@ -60,25 +61,24 @@
...
-  private Double fakeDelayMovingAverageSamples;
+  private final Integer maxNumberOfKeys;
 
-  @Value("${services.submission.retention-days}")
-  private Integer retentionDays;
+  private Double fakeDelay;
 
-  @Value("${services.submission.payload.max-number-of-keys}")
-  private Integer maxNumberOfKeys;
+  @Autowired
```
jwedel, line 72: As mentioned before, `@Autowired` is not required at constructors when there is only one constructor.

pithumke:
Hi @indyaah,
#240 has been merged and #272 is on its way to master.

services/submission/src/main/java/app/coronawarn/server/services/submission/config/SubmissionServiceConfig.java
```diff
@@ -0,0 +1,64 @@
...
+
+@Component
+@ConfigurationProperties(prefix = "services.submission")
+public class SubmissionServiceConfig {
+
+  // Exponential moving average of the last N real request durations (in ms), where
+  // N = fakeDelayMovingAverageSamples.
+  private Double initialFakeDelayMilliseconds;
+  private Double fakeDelayMovingAverageSamples;
+  private Integer retentionDays;
```
elch78, line 15: I would suggest to use java.time.Duration instead of int. Spring Boot should convert the input automatically. Using Duration in DiagnosisKey would make the implementation easier. Something like
`return Duration.between(Instant.ofEpochMilli(rollingStartNumber), Instant.now())
      .compareTo(durationToRetain) <= 0;`

pithumke: Hey @elch78,
thanks for the feedback! Yes I guess you are right and we should use domain specific types where ever possible. However, please understand that the purpose of this PR is to "only" remove the `@Value` annotations and not to much more refactoring above that. Since this PR is already clocking in at almost 1000 LOC delta, I kindly ask you to create an issue for this so we can pick it up in a separate (more manageable) PR. Thanks! :)

pithumke:
Closes #207

pithumke:
Big thank you to all community contributors, you rock! 🚀 ⭐

services/submission/src/main/java/app/coronawarn/server/services/submission/config/SubmissionServiceConfig.java
elch78: Agree

Issue #205, opened 2020-05-20
Update signature algorithm to SHA-256 + ECDSA P-256


pithumke:
## Feature description
The current CWA server implementation signs diagnosis key packages with the ``Ed25519`` algorithm. However, the current [Google/Apple spec](https://developer.apple.com/documentation/exposurenotification/setting_up_an_exposure_notification_server) demands ``SHA-256`` hash with ``ECDSA P-256`` signature ([OID 1.2.840.10045.4.3.2](http://oid-info.com/get/1.2.840.10045.4.3.2))

PR #240, opened 2020-05-21
Update signature info

pithumke:
Latest information is that we will use:

| Property  | Value |
| ------------- | ------------- |
| App Bundle ID | `de.rki.coronawarnapp`  |
| Android Package | `de.rki.coronawarnapp`  |
| Signature Algorithm (OID) | `1.2.840.10045.4.3.2`  |
| Verification Key ID | To be defined  |
| Verification Key Version | To be defined  |

Caveats of this PR:
- Provided signature info does not match the actual signature algorithm currently used for signing.
  - Issue: #205 
  - Code: [AbstractSigningDecorator.java#L40](https://github.com/corona-warn-app/cwa-server/blob/master/services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/structure/archive/decorator/signing/AbstractSigningDecorator.java#L40)

Linked issues:
- #183
- #207
- #238


christian-kirschnick:
Very cool, thanks Pit!

PR #244, opened 2020-05-22
Added myself and Timon to CODEOWNERS

stevesap:


stevesap:
i'm joined to sap org now

PR #252, opened 2020-05-22
Update SECURITY.md

tkowark:
Updated instructions for product and versions selection in vulnerability reporting


stevesap:
@tkowark I think it would be more easily understood how to find the issue submission tool if we link directly to https://www.sap.com/about/trust-center/incident-management.html and ask they report a security issue under Security Researchers.

PR #253, opened 2020-05-22
Documenting docker-compose limitation

ole-lilienthal:


ole-lilienthal:
@jakobmoellersap Please cross-read if this would have helped

README.md
```diff
@@ -63,6 +63,10 @@ postgres      | A [postgres] database installation
 pgadmin       | A [pgadmin](https://www.pgadmin.org/) installation for the postgres database                | http://localhost:8002 <br> Username: user@domain.com <br> Password: password
 cloudserver   | [Zenko CloudServer] is a S3-compliant object store  | http://localhost:8003/ <br> Access key: accessKey1 <br> Secret key: verySecretKey1
 
+##### Known Limitation
+
+The docker-compose runs into a timing issue in some cases when the create-bucket target runs before the objectstore is available. The mitigation is easy: after running ```docker-compose up``` wait until all components are initialized and running. Afterwards, trigger the ```create-bucket``` service manually by running ```docker-compose run create-bucket```. If you want to trigger distribution runs, run ```docker-compose run distribution```. The timing issue will be fixed in a future release.
```
jakobmoellersap, line 68: IMO this is almost always the case. I never had a scenario where docker-compose ran through at once without issue. Maybe it makes sense to remove the dependencies and just document an order of execution.

I agree that this fixes the issue for me. However, "docker-compose up" from scratch still is basically creating errors in the aws-sdk without it being necessary and creating longer developer stories than necessary. 

If this is considered acceptable in this release (since it is easily avoidable) until the mentioned future release, this is okay. I will mention it being a big issue-starter though if people run into this without realising the effect of this limitation.

ole-lilienthal: That seems to be depending on the machine it is run on. I tested it with three additional colleagues before pushing without an issue. To my knowledge there is no way to include the component in docker-compose but stop it from running on docker compose up. Otherwise I would have agreed. I hope we get this fixed in general rather soonish, the idea how it can be fixed is already existing.

ole-lilienthal:
Please also see Issue #255

PR #259, opened 2020-05-22
use maven with --batch-mode

CCFenner:
Maven should run in *batch mode* to avoid extensive build logs and ease readability. This will omit the *progress* in the build log:

```log
Downloading from central: https://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-plugins/12/maven-plugins-12.pom
Progress (1): 2.2/12 kB
Progress (1): 5.0/12 kB
Progress (1): 7.8/12 kB
Progress (1): 11/12 kB 
Progress (1): 12 kB   
```

Also the use of `-Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn` could be considered to remove the `"Downloading from ..."` entries completely.


Bananeweizen:
This change looks good to me.

Regarding the use of the logger workaround, there is a better alternative: Require at least Maven 3.6.1 via maven-enforcer-plugin and use the command line option --no-transfer-progress, which is available since Maven 3.6.1.

Issue #78, opened 2020-05-12
Add mobile configuration


christian-kirschnick:
Mobile needs to be able to fetch additional configuration/settings from the server. The process should be similar to the exposure configuration.

Therefore, add a new app-config.yaml to the codebase. The app-config should contain

```yaml
risk_classes: 
  - 
    class: LOW
    max: 3
    min: 1
    uri: "https://www..."
  - 
    class: MID
    max: 5
    min: 3
    uri: "https://www..."
  - 
    class: HIGH
    max: 8
    min: 5
    uri: "https://www..."
```
We need validations & tests (also similar to exposure config). The YAML needs to be parsed to the proto message, so it can used in the distribution service for pushing to CDN.

PR #260, opened 2020-05-22
Fix/Validate exposure configuration at runtime

### LINKED ISSUE #122: Exposure configuration parameters not validated
The exposure configuration parameters are currently not validated after being read from the respective yml file. The validation code in ``ExposureConfigurationValidator`` is currently used in tests only.

However, the expected behavior shall be the following:
 - The ``RiskScoreParameters`` object is constructed based on yml file. (✅)
 - ``ExposureConfigurationValidator`` performs validation of the parameters.
 - In case of a failed validation, **do not** publish the exposure configuration parameters, **but do** generate and publish the diagnosis key bundles.
 - Clarify whether to keep any previously present (valid) configuration parameters on the object store if a validation error occurs.
 - Any validation errors shall be logged appropriately.

michael-burwig:
Fixes #122 

- Skips publishing the exposure configuration if its validation fails.
- Changes parameters directory name from ``parameters`` to``configuration``.
- Changes the name of the exposure configuration archive from ``index`` to ``exposure_configuration``
- Applies refactoring to ease the support of other configuration files. (see #78)


services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/appconfig/structure/directory/AppConfigurationDirectory.java
```diff
@@ -0,0 +1,96 @@
...
+    this.addWritable(new IndexingDecoratorOnDisk<>(countryDirectory));
+  }
+
+  private void addExposureConfigurationIfValid() {
+    try {
+      RiskScoreParameters exposureConfig = ExposureConfigurationProvider.readMasterFile();
+      AppConfigurationValidator validator = new ExposureConfigurationValidator(exposureConfig);
+      addArchiveIfMessageValid(EXPOSURE_CONFIGURATION_FILE_NAME, exposureConfig, validator);
+    } catch (UnableToLoadFileException e) {
+      logger.error("Exposure configuration will not be published! Unable to read configuration file from disk.");
```
christian-kirschnick, line 76: Should add the exception to the log entry as well.

Issue #136, opened 2020-05-17
Make DistributionKeyService I/O non-blocking


pithumke:
## Current Implementation
The `DistributionKeyService` I/O (e.g. `saveDiagnosisKeys`, `getDiagnosisKeys`, `applyRetentionPolicy`) is thread blocking.

## Suggested Enhancement
Make the `DistributionKeyService` I/O non-blocking through the use of [Futures](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Future.html) or other suitable constructs. Adapt [SubmissionController](https://github.com/corona-warn-app/cwa-server/blob/master/services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java) and [Distribution runners](https://github.com/corona-warn-app/cwa-server/tree/master/services/distribution/src/main/java/app/coronawarn/server/services/distribution/runner) accordingly.

## Expected Benefits
Performance improvement / lower resource footprint.

ebratter:
Per recent discussions with @christian-kirschnick it was agreed to close this issue due to current performance results.

PR #263, opened 2020-05-22
Made DistributionKeyService async based on CompletableFuture (#136)

dennis-menge:
DistributionKeyService is now based on CompletableFuture
Also migrated SubmissionController from DeferredResult to CompletableFuture

Feedback is highly welcome :)


johanneseschrig:
Hi @dennis-menge! Thanks for you contribution! We had some discussions on the linked issue and we have decided to do some more performance testing before introducing these additional complexities at this point. Apologies for not communicating this more clearly in the linked issue, we will keep this PR in the back of our heads if performance testing should deem this necessary. Thanks!

dennis-menge:
Alright, good to know. 

At this point, you could also think about changing from MVC to Webflux which would probably make more sense, and would not be that much effort in the current state of the implementation.

PR #264, opened 2020-05-22
Adds Risk Score Classification Configuration

### LINKED ISSUE #78: Add mobile configuration

michael-burwig:
Resolves #78 

- Adds proto files and master.yml for risk score classification.
- Adds ``risk_score_classification`` archive to ``configuration`` directory.
- Adds validation code + tests for risk score classifications.
- Validates risk score classification master file during test runs.

### DIFF
--- a/services/distribution/src/main/resources/risk-score-classification/master.yaml
+++ b/services/distribution/src/main/resources/risk-score-classification/master.yaml
@@ -1,0 +1,26 @@
+# This is the Risk Score Classification master file which partitions the risk
+# score value range (0-255) into distinct risk classes that will be used to by
+# the app in order to present the correct risk classification in a user
+# friendly manner, based on the underlying total risk score, to the user.
+#
+# The risk classes must not overlap and
+# cover the full risk score value range (0-255).
+#
+# Change this file with caution!
+
+risk_classes:
+  -
+    class: LOW
+    min: 0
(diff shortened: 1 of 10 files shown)


services/distribution/src/main/resources/risk-score-classification/master.yaml
```diff
@@ -0,0 +1,27 @@
...
+# cover the full risk score value range (0-255).
+#
+# Change this file with caution!
+
+risk_classes:
+  -
+    label: "LOW"
+    min: 0
+    max: 100
+    url: "https://www..."
```
christian-kirschnick, line 16: Let's include default URI's to coronawarn.app for now, not broken ones.

PR #265, opened 2020-05-23
reduce version definitions in POMs

CCFenner:
The PR removes the `<version>` tag from each POM where it is not necessary.

**Changes**:
- use `${project.version}` for dependency versions
- inherit `<version>` where possible
- define parent `<version>` via inherited property `${revision}`

**Question:** Is the `flatten-maven-plugin` needed as documented [here](https://maven.apache.org/maven-ci-friendly.html#install-deploy)?

The version definition locations can even be reduced further to one by using a `.mvn/maven.config` file defining `revision`.


johanneseschrig:
Hi @CCFenner, have you tried introducing a `.mvn/maven.config` file defining `revision`? If that reduces the places we need to make manual changes, maybe we can implement that directly in this PR? Thanks!

CCFenner:
Yes, that works, will update the PR.

CCFenner:
CircleCI still has *unauthorized*. issue, at least when I trigger it via commit/rebase.

PR #276, opened 2020-05-23
Update spring-boot-starter-parent to 2.3.0-RELEASE

MKusber:
This PR will update some of our dependencies to a more recent version, reducing attack surface.


stevesap:
Local clean build also passes

PR #283, opened 2020-05-24
Fix/editorconfig clashes w checkstyle

michael-burwig:
- indent_size entry in seemed to be clashing with certain checkstyle.xml entries
- removes/adds a number of of whitespaces/newlines

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,7 +2,6 @@ root = true
 
 [*]
 indent_style = space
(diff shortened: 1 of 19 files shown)


CCFenner:
You can only remove the indent_size for java files, this way it would be still valid for `xml`, etc.

```
[*]
...

[!*.java]
indent_size = 2
```

Issue #255, opened 2020-05-22
Docker-compose limitation


ole-lilienthal:
## Current Implementation
As documented in the [README](https://github.com/corona-warn-app/cwa-server#known-limitation) the docker-compose currently has limitations. When running ```docker-compose up``` the create-bucket target is not waiting until the objectstorage is available and the distribution service is not waiting until the bucket was created.

## Suggested Enhancement
The docker compose should be enhanced by one of the proposals provided by the official documentation: https://docs.docker.com/compose/startup-order/.

* The create-bucket should wait until the objectstore is available.
* The distribution service should wait until the bucket was successfully created.

PR #285, opened 2020-05-24
Fix/docker compose limitation #255

AlexanderBarbie:
I added the wait-for-it script from https://github.com/vishnubob/wait-for-it as suggested in  https://docs.docker.com/compose/startup-order/ and adjusted the docker-compose file.

The timeout with 30 seconds is working for me on two different systems. With a timeout of "only"  15 seconds the container still crashes, since the object store is not up until then.

This should fix your issue #255 Docker-compose limitation @ole-lilienthal .


docker-compose.yml
```diff
@@ -83,7 +83,10 @@ services:
     environment:
       - AWS_ACCESS_KEY_ID=${OBJECTSTORE_ACCESSKEY}
       - AWS_SECRET_ACCESS_KEY=${OBJECTSTORE_SECRETKEY}
-    command: s3api create-bucket --bucket cwa --endpoint-url http://objectstore:8000 --acl public-read
+    entrypoint: ["/root/scripts/wait-for-it/wait-for-it.sh", "http://objectstore:8000", "-q","-t", "30", "--"]
```
ole-lilienthal, line 86: Shouldn't this be "objectstore:8000" instead of "http://objectstore:8080"??
I would also prefer to have it not running silently, thus to remove -q

AlexanderBarbie: I put it in silent mode, due to the misleading output:

create-bucket_1  | wait-for-it.sh: waiting 30 seconds for objectstore:8000
create-bucket_1  | wait-for-it.sh: timeout occurred after waiting 30 seconds for objectstore:8000

Furthermore, I made the experience in another project that I had to wait for the service at the address to be available, instead of the container and since aws calls http://objectstore:8000 I thought this could be the case here, too. But I tested in with "objectstore:8000" and this works just fine, too.

scripts/wait-for-it/LICENSE
```diff
@@ -0,0 +1,20 @@
+The MIT License (MIT)
```
stevesap, line 1: Hi @AlexanderBarbie, could you please move this license into our dedicated [THIRD-PARTY-NOTICES file](https://github.com/corona-warn-app/cwa-server/blob/master/THIRD-PARTY-NOTICES)? Thanks.

docker-compose.yml
AlexanderBarbie: I am trying to find the part with the distribution service that has to be fixed. If I run docker-compose up the distribution service starts and after the bucket was created it just exchanges stuff and exits with code 0. Did this a few times now and always the same. Am I missing something?

ole-lilienthal: From my perspective the output wasn't misleading - should be fine after you changed it to objectsore:8080, isn't it?
On some machines the distribution service runs before the bucket was created. That's only the case on the first run. In theory the distribution run should only happen after the bucket was created. But that's more tricky to tackle and I just wanted to understand if you have any idea in that regards.

AlexanderBarbie: i am trying to reproduce this, but on my machine the distribution service takes a few seconds to the application and the bucket ist created before. I only get a crashing distribution service, if the bucket is missing, hence, if the bucket container crashes before the objectstore is reachable. My first idea would be to ping the create_bucket container until it stopped and start the distribution service afterwards. However, this should be a separate  issue then, since this maybe needs adjustments in the Dockerfile.

ole-lilienthal:
@michael-burwig Please have another look from a License perspective. Functional wise it's fine for me. 
Will create a second ticket for the distribution service and think about a possible readme adjustment.

THIRD-PARTY-NOTICES
```diff
@@ -680,6 +685,27 @@ so, subject to the following conditions:
...
 
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+--------------------------------------------------------------------------------
+The MIT License (MIT)
```
michael-burwig, line 696: Please, follow the pattern:
```
--------------------------------------------------------------------------------
<LICENSE_NAME> (<COMPONENT_NAME>)
\n
<LICENSE_TEXT>
--------------------------------------------------------------------------------
```

Other than that, this looks good to me.

AlexanderBarbie:
oh, that was not planned

PR #288, opened 2020-05-24
Resolved a couple of sonar code smells

stevesap:
- Invoke method(s) only conditionally. Use the built-in formatting to construct this argument
- Format specifiers should be used instead of string concatenation.
- Remove unnecessary or unused import
- Add a private constructor to hide the implicit public one
- Invalid symbol in JavaDoc


services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/diagnosiskeys/util/DateTime.java
```diff
@@ -33,6 +33,10 @@
  */
 public class DateTime {
 
+  private DateTime() {
+    throw new IllegalStateException("DateTime class");
```
michael-burwig, line 37: .

Issue #270, opened 2020-05-23
Sonar - Exception handling


johanneseschrig:
SonarCloud is [currently complaining](https://sonarcloud.io/project/issues?id=corona-warn-app_cwa-server&resolved=false&severities=MAJOR) about exception handling in many places. We need to mitigate/audit these issues.

elch78:
Fixed the "dont log and throw" issues.

In my projects I normally disable this rule: "Define and throw a dedicated exception instead of using a generic one." As long as there's no way to handle the exception separately it's totally fine to throw RuntimeException.

stevesap:
The remaining instances of 'Define and throw a dedicated exception instead of using a generic one' have been marked as wontfix in sonar where no less-generic alternative to RuntimeException was found to be suitable.

PR #293, opened 2020-05-24
Only throw exception, don't log (#270)

elch78:


stevesap:
FYI - Relates to Sonar rule: [Exceptions should be either logged or rethrown but not both](https://rules.sonarsource.com/java/tag/error-handling/RSPEC-2139)

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/component/CryptoProvider.java
```diff
@@ -98,8 +98,7 @@ public PrivateKey getPrivateKey() {
       try (InputStream privateKeyStream = privateKeyResource.getInputStream()) {
         privateKey = getPrivateKeyFromStream(privateKeyStream);
       } catch (IOException e) {
-        logger.error("Failed to load private key from {}", privateKeyPath, e);
-        throw new RuntimeException(e);
+        throw new RuntimeException("Failed to load private key from " + privateKeyPath, e);
```
stevesap, line 101: We propose the less generic `UncheckedIOException` which extends `RuntimeException` and also wraps an `IOException`

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/component/CryptoProvider.java
```diff
@@ -114,8 +113,7 @@ public Certificate getCertificate() {
       try (InputStream certStream = certResource.getInputStream()) {
         this.certificate = getCertificateFromStream(certStream);
       } catch (IOException | CertificateException e) {
-        logger.error("Failed to load certificate from {}", certificatePath, e);
-        throw new RuntimeException(e);
+        throw new RuntimeException("Failed to load certificate from " + certificatePath, e);
```
stevesap, line 116: For handling the `IOException`, we propose the less generic `UncheckedIOException` which extends `RuntimeException` and also wraps an `IOException`

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/io/IO.java
```diff
@@ -48,8 +48,7 @@ public static void makeNewFile(File root, String name) {
         throw new IOException("Could not create " + name + ", file already exists");
       }
     } catch (IOException e) {
-      logger.error("Failed to create file: {}", name, e);
-      throw new RuntimeException(e);
+      throw new RuntimeException("Failed to create file: " + name, e);
```
stevesap, line 51: We propose the less generic `UncheckedIOException` which extends `RuntimeException` and also wraps an `IOException`

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/io/IO.java
```diff
@@ -63,8 +62,7 @@ public static void writeBytesToFile(byte[] bytes, File outputFile) {
     try (FileOutputStream outputFileStream = new FileOutputStream(outputFile)) {
       outputFileStream.write(bytes);
     } catch (IOException e) {
-      logger.error("Could not write file {}", outputFile);
-      throw new RuntimeException(e);
+      throw new RuntimeException("Could not write file " + outputFile, e);
```
stevesap, line 65: We propose the less generic `UncheckedIOException` which extends `RuntimeException` and also wraps an `IOException`

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/component/CryptoProvider.java
elch78: UncheckedIOException is not compatible here. Is it likely that you want to distinguish between RuntimeException and UncheckedIOException? If not I would say it only makes the code unnecessarily more complex. For the cases where UncheckedIOException is compatible I also don't any benefit of the more specific exception.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/component/CryptoProvider.java
elch78: Replaced with UncheckedIOException.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/component/CryptoProvider.java
elch78: Replaced with UncheckedIOException and RuntimeException in case of CertificateException.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/io/IO.java
elch78: Replaced with UncheckedIOException.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/io/IO.java
elch78: Replaced with UncheckedIOException.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/component/CryptoProvider.java
michael-burwig: Just checked with Steve and we agree that sticking with the initial ``} catch (IOException | CertificateException e) {`` works fine for us here.

If you don't mind reverting this particular change, we can go ahead and merge.

elch78: Sure

PR #296, opened 2020-05-25
Ensure StopWatch is stopped

espiegelberg:
Wrap the StopWatch.stop() in a finally() block so as to ensure it is stopped in the (unlikely but possible) scenario that an exception is thrown.


services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java
```diff
@@ -116,17 +116,24 @@ private void setRealDeferredResult(DeferredResult<ResponseEntity<Void>> deferred
     forkJoinPool.submit(() -> {
```
michael-burwig, line 116: Hi @espiegelberg,
good find! After consulting with my colleagues, I am suggesting the following:
```
    forkJoinPool.submit(() -> {
      StopWatch stopWatch = new StopWatch();
      stopWatch.start();
      if (!this.tanVerifier.verifyTan(tan)) {
        deferredResult.setResult(buildTanInvalidResponseEntity());
      } else {
        try {
          persistDiagnosisKeysPayload(exposureKeys);
          deferredResult.setResult(buildSuccessResponseEntity());
          stopWatch.stop();
          updateFakeDelay(stopWatch.getTotalTimeMillis());
        } catch (Exception e) {
          deferredResult.setErrorResult(e);
        }
      }
    });
```

This way, we only update the fake request delay after data is actually written to the database. As a bonus, we don't have to worry about stopping the ``StopWatch``.

Thanks for spending time on this!

Cheers,
Michael

stevesap:
@espiegelberg would you be able to update as per the suggestion? Would be nice to have this soon if possible

services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java
espiegelberg: I have made the update to match your suggestion and committed and pushed.

Having taken a closer look at the logic though, it appears the StopWatch is never stopped if the line "if (!this.tanVerifier.verifyTan(tan)) {" is true and buildTanInvalidResponseEntity() is invoked. I assume this is an oversight, correct? If yes, the StopWatch should be stopped after the deferredResult's set.

For what it's worth, I also recommend that the StopWatch be stopped somewhere in a finally block, as an unintended Exception could also prevent stoppage as well.

I am not sure if not explicitly stopping a StopWatch has unintended consequences or not; I assume explicitly stopping it is considered a best practice (similar to closing a resource).

stevesap:
Thanks @espiegelberg for the prompt update!

services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java
michael-burwig: Hi @espiegelberg,
from looking at the implementation of ``StopWatch``, stopping it only matters when you are accessing its time measurement (which we are not doing in the exception case now).

``StopWatch`` is simply a neat wrapper for
```
starttime = System.nanoTime();
...
endtime = System.nanoTime();
```
Based on that there I do not expect any leakage here and ultimately see it as a judgment call whether to do it or not.

espiegelberg: +1

PR #299, opened 2020-05-25
Adjust Risk Score Parameters (#298)

### LINKED ISSUE #298: RKI: Adjust Risk Score Parameters in master config
RKI decided for the following risk score parameter values:

```
1       2       3       4       5       6       7       8 // Bucket No.
----------------------------------------------------------
0       1       2       4       5       7       8       8 // Duration
0       0       1       2       5       7       8       8 // Attenuation
1       1       1       1       2       6       8       6 // Days
1       2       3       4       5       6       7       8 // Transmission Risk 
```

max-ae:
Closes #298


stevesap:
@christian-kirschnick do we also need to update https://github.com/corona-warn-app/cwa-server/blob/master/common/protocols/src/main/proto/app/coronawarn/server/common/protocols/internal/risk_score_parameters.proto or will the master.yaml values overwrite these?

christian-kirschnick:
@stevesap This PR just changes the master config for the values, so we don't need any changes in the proto files. OK from my side.

stevesap:
```
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.101 s <<< FAILURE! - in app.coronawarn.server.services.distribution.assembly.appconfig.ExposureConfigurationProviderMasterFileTest
[ERROR] testMasterFile  Time elapsed: 0.1 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: 

Expecting:
 <[RiskLevelValidationError{parameter='attenuation', riskLevel='gt63Le73Dbm'}, RiskLevelValidationError{parameter='attenuation', riskLevel='gt73Dbm'}, RiskLevelValidationError{parameter='duration', riskLevel='eq0Min'}]>
to be equal to:
 <[]>
but was not.
```

Seems the validator doesn't like some of the new values?

max-ae:
The '0' values for Days and Attenuation seem to be the issue
https://github.com/corona-warn-app/cwa-server/blob/4a35477e89099e228d0ddaa6501c4f09f8beda57/services/distribution/src/main/resources/exposure-config/master.yaml#L5

stevesap:
@max-ae as we have now confirmed these values with the RKI, could you also adjust the validation to support 0 values?

stevesap:
@max-ae please hold off on change. I'm getting this double checked and will update shortly.

max-ae:
> @max-ae please hold off on change. I'm getting this double checked and will update shortly.

Okay 👍🏼

stevesap:
@max-ae we've confirmed 0-8 value range is correct so if you can go ahead and adjust the validation + unit tests that would be great.

max-ae:
@stevesap I adjusted the validation, but could you or someone else help fix the unit tests? 
I think the following file should be adjusted too, so that a value of 0 does not equal UNSPECIFIED
https://github.com/corona-warn-app/cwa-server/blob/9889d29313184499b40093ff846ffaebbe04d7ff/common/protocols/src/main/proto/app/coronawarn/server/common/protocols/internal/risk_level.proto#L7

max-ae:
@stevesap With the latest commit, the tests now run, but the validator allows '0' values for the transmission parameter too. Is this intended? Or should these stay in between 1-8?
Also, if a value in the configuration file is missing, (see https://github.com/corona-warn-app/cwa-server/blob/9889d29313184499b40093ff846ffaebbe04d7ff/services/distribution/src/test/resources/parameters/score_negative.yaml#L9) it will get read as RiskLevel.UNSPECIFIED, however, the validator no longer throws an error for this case as `0` values should be allowed now. Is this also intended?

stevesap:
@max-ae please hold - i'm getting some help on this. Thanks for your patience.

stevesap:
@max-ae sorry for the delay. So your approach so far is correct.

Transmission parameter valid range is 1-8 while for other three it is 0-8.

For now, please update to ignore the ``score_negative.yaml`` test and we will create an enhancement ticket and link here to validate that there are no missing yaml properties on load of the yaml into the proto objects.

max-ae:
> For now, please update to ignore the `score_negative.yaml` test and we will create an enhancement ticket and link here to validate that there are no missing yaml properties on load of the yaml into the proto objects.

Done 👍🏼

services/distribution/src/test/java/app/coronawarn/server/services/distribution/assembly/appconfig/validation/ExposureConfigurationValidatorTest.java
```diff
@@ -72,7 +72,9 @@ private ValidationResult getResultForTest(TestWithExpectedResult test)
   private static Stream<Arguments> createFailedTests() {
     return Stream.of(
         ScoreTooHigh(),
-        ScoreNegative(),
+        // TODO adjust test when new validation for yaml-Files is implemented (check that no attributes are missing
```
stevesap, line 75: I've created the ticket now so please update to:
// TODO cwa-server/#320 Validate that no attributes are missing in .yaml

max-ae: Updated

Issue #306, opened 2020-05-25
CryptoProvider is not checking resources on startup


elch78:
## Current Implementation
Resource are loaded in the getPrivateKey and getCertificate methods. If the resources can't be loaded an exception is thrown at runtime. 

Also the UnitTest CryptoProviderTest is not testing what the name 'constructorInitializesCryptoArtifacts' suggests. 

## Suggested Enhancement
Load the resources in a @PostConstruct method. If resources can't be loaded an exception is thrown at startup and the application will not start.

## Expected Benefits
Configuration errors are detected on startup not during runtime. If the application is deployed in a blue/green fashion the error can be fixed before the user is affected.

WilliamLin:
Hi, I am new to the open-source project, but I would like to start making the contribution. This task seems simple enough for a beginner. Please assign this task to me, or any other low hanging fruit.  Thanks

PR #307, opened 2020-05-25
Crypto provider load resources on startup (#306)

elch78:
This pullrequest is based on the branch for issue #270


services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/component/CryptoProvider.java
```diff
@@ -65,6 +66,15 @@
     Security.addProvider(new BouncyCastleProvider());
   }
 
+  /**
+   * Reads the {@link Certificate} and {@link PrivateKey} configured in the application properties.
+   */
+  @PostConstruct
```
stevesap, line 72: Hi @elch78, since we are using Constructor Injection, the loading of PrivateKey and Certificate should be already possible in the constructor as the privateKeyStream dependency is injected and available. Any particular reason to do it in PostConstruct?

elch78: Hi @stevesap . The reason to  not do it in the constructor was just the habit to avoid IO in the constructor because the constructor is normally not mockable. But since this whole class is only an abstracting for the disc IO I also think that the code can be called from the constructor. I'll change that. Then also the members are obsolete and the loaded member can be final.

PR #308, opened 2020-05-25
Validate submission payload

### LINKED ISSUE #179: Apple's Proposal for Exposure Key Validation
#92  Current Implementation

We used sensible assumptions for key validation.

## Suggested Enhancement

[Apple proposes](https://developer.apple.com/documentation/exposurenotification/setting_up_an_exposure_notification_server?changes=latest_beta): 

> An EN server must reject invalid key files uploaded from a client app. Uploaded key data is considered invalid if:
>
> Any ENIntervalNumber values from the same user are not unique
> 
> There are any gaps in the ENIntervalNumber values for a user
> 
> Any keys in the file have overlapping time windows
> 
> The period of time covered by the data file exceeds 14 days
> 
> The TEKRollingPeriod value is not 144
> 
> You may optionally want to validate the clock time the device submits.

## Expected Benefits

Adherence to reference implementations.

michael-burwig:
Resolves #179 
Introduces the class``SubmissionPayloadValidator`` that validates the following contraints on the request body of diagnosis key submission requests:
- StartInvervalNumber values from the same user shall be unique.
- There must be no gaps for StartInvervalNumber values for a user.
- There must not be any keys in the file have overlapping time windows.
- The period of time covered by the data file must not exceed 14 days.

Changes the ``DiagnosisKey`` validation logic to only accept ``144`` as a valid value for``RollingPeriod``. The methods in ``DiagnosisKeyBuilder`` are adjusted so that ``RollingPeriod`` is now an optional parameter that defaults to ``144`` if not specified.


common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/DiagnosisKey.java
```diff
@@ -58,7 +67,8 @@
   @ValidRollingStartIntervalNumber
   private int rollingStartIntervalNumber;
 
-  @Min(value = 1, message = "Rolling period must be greater than 0.")
+  @Min(value = EXPECTED_ROLLING_PERIOD, message = "Rolling period must be " + EXPECTED_ROLLING_PERIOD + ".")
+  @Max(value = EXPECTED_ROLLING_PERIOD, message = "Rolling period must be " + EXPECTED_ROLLING_PERIOD + ".")
```
johanneseschrig, line 71: I guess `@Range` might save one of these.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/runner/TestDataGeneration.java
```diff
@@ -182,8 +181,8 @@ private DiagnosisKey generateDiagnosisKey(long submissionTimestamp) {
   }
 
   /**
-   * Returns a random rolling start number (timestamp since when a key was active, represented by a 10 minute interval
-   * counter.) between a specific submission timestamp and the beginning of the retention period.
+   * Returns a random rolling start interval number (timestamp since when a key was active, represented by a 10 minute
```
johanneseschrig, line 184: Not introduced by this PR, I know. But should we really be using random values in tests? Shouldn't we refactor this to a parameterized test with some values, and edge cases specifically? Open issue would be fine by me.

PR #311, opened 2020-05-26
WIP: replace size-comparison in test with "hasSameSizeAs"

nils4cosee:
My SonarLint-plugin displayed this as warning and I just applied the quick-fix.


michael-burwig:
Hi @nils4cosee,

thanks for the quick fix. It looks like there is similar occurrence in ``DiagnosisKeyServiceTestHelper``:
```
assertThat(actKeys).withFailMessage("Cardinality mismatch").isNotNull().hasSize(expKeys.size());
```

nils4cosee:
@michael-burwig I fixed that warning as well.

nils4cosee:
I was a bit quick with creating this PR. Please do not merge at the moment. Even though it is only a small fix. You can do the fix yourself if you want, but please not (yet) with my name on it.

Issue #315, opened 2020-05-26
Distribution Service: Provide Metrics to Prometheus


johanneseschrig:
Consider prometheus pushgateway.

johanneseschrig:
Wrong issue linked

PR #316, opened 2020-05-26
Submission Service: Actuator Metric Exposure for Prometheus

johanneseschrig:
This will expose the `/actuator` endpoint on port 8081. This port will not be publicly reachable, but only Prometheus will have access to poll (to be configured in PaaS). Had to allow `GET` in the Spring Firewall, however security config will return 403 if any GET is performed on 8080. 

Resolves https://github.com/corona-warn-app/cwa-server/issues/315

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ The docker-compose contains the following services:
 
 Service       | Description | Endpoint and Default Credentials
 --------------|-------------|-----------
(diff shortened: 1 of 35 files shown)


MKusber:
As discussed. Please consider syncing with operations and monitoring wrt. adding auth to the endpoint, but for now, this looks good.

PR #319, opened 2020-05-26
Improve header handling

MKusber:
This PR changes header validation behavior:

* `cwa-fake` will be parsed as a String, preventing `NumberFormatException`s that would log this value
* Adjust logging of unverified TANs to agreed upon string.


johanneseschrig:
Please fix tests.

services/submission/src/main/java/app/coronawarn/server/services/submission/verification/TanVerifier.java
```diff
@@ -70,7 +70,7 @@ public boolean verifyTan(String tan) {
     String trimmedTan = tan.trim();
 
     if (!checkTanSyntax(trimmedTan)) {
-      logger.debug("TAN Syntax check failed for TAN: {}", trimmedTan);
+      logger.debug("TAN Syntax check failed for TAN: {}", trimmedTan.substring(0,10));
```
michael-burwig, line 73: Looks like it could be vulnerable to an ``IndexOutOfBoundsException``

MKusber:
* cwa-fake will be parsed as a String, preventing NumberFormatExceptions that would log this value
* Adjust logging of unverified TANs to agreed upon string.

PR #346, opened 2020-05-27
Add S3 Retention Period

### LINKED ISSUE #329: Retention: Enforce period on Object Store
Currently, the retention period is only enforced for keys on the database, but it should be enforced on the S3 as well.

Therefore, we need to delete the items on the S3, when the retention policy is reached.

When the retention mechanism is kicked off, and then delete the following keys with the prefix:
`version/v1/diagnosis-keys/country/DE/date/<RETENTIONDATE>`
Where RETENTIONDATE is `current date - retention period`.

Ideally, we should also try to delete older items as well, in case the retention run did not succeed or did not run at all.

UnchartedBull:
This PR adds an S3 retention run to the Retention Runner. The S3 retention run makes sure, that there are no files older than `retentionDays` on the S3 compatible storage.

Closes #329.


johanneseschrig:
Please add tests and review the Sonar quality gate.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/runner/RetentionPolicy.java
```diff
@@ -46,27 +46,32 @@
...
+  private final S3Distribution s3Distribution;
+
   /**
    * Creates a new RetentionPolicy.
    */
   RetentionPolicy(DiagnosisKeyService diagnosisKeyService,
       ApplicationContext applicationContext,
-      DistributionServiceConfig distributionServiceConfig) {
+      DistributionServiceConfig distributionServiceConfig,
+                         S3Distribution s3Distribution) {
```
michael-burwig, line 57: formatting issue

services/distribution/src/main/java/app/coronawarn/server/services/distribution/runner/S3Distribution.java
```diff
@@ -63,4 +82,39 @@ public void run(ApplicationArguments args) {
...
+
+    diagnosisKeysObjects.stream()
+        .filter(diagnosisKeysObject -> {
+          Matcher matcher = pattern.matcher(diagnosisKeysObject.getObjectName());
+          return matcher.matches() && LocalDate.parse(matcher.group(1), DateTimeFormatter.ISO_LOCAL_DATE)
+              .isBefore(cutOffDate);
+        })
+        .map(diagnosisKeysObject -> {
+          try {
+            return objectStoreAccess.deleteObjectsWithPrefix(diagnosisKeysObject.getObjectName());
```
michael-burwig, line 110: We could consider to not make ``deleteObjectsWithPrefix`` return this list but throw an Exception right away. As far as I can see, the result is not used anywhere else (maybe in tests).

PR #358, opened 2020-05-28
Create pull_request_template.md

tkowark:
Pull Request Template containing the main points that should be taken into account when submitting a new request.


.github/pull_request_template.md
```diff
@@ -0,0 +1,17 @@
+<!-- 
+Thank you for supporting us with your Pull Request! 🙌 ❤️  
+Before submitting, please take the time to check the points below and provide some descriptive information.
+-->
+
+## Checklist
+
+* [ ] Test your changes as thoroughly as possible before you commit them. Preferably, automate your test by unit/integration tests.
+* [ ] If this PR comes from a fork, please [Allow edits from maintainers](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)
+* [ ] Set a speaking title. Format: {task_name} (closes #{issue_number}). For example: Use logger (closes #41)
```
stevesap, line 10: I would say 'Set a title' sounds a bit better

johanneseschrig: But the focus here is not on setting a title, but choosing one that is descriptive. So that should in some way be reflected.

.github/pull_request_template.md
```diff
@@ -0,0 +1,17 @@
...
+* [ ] [Link your Pull Request to an issue](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) (if applicable)
+* [ ] Create Work In Progress [WIP] pull requests only if you need clarification or an explicit review before you can continue your work item.
+* [ ] Make sure that your PR is not introducing _unncessary_ reformatting (e.g., introduced by on-save hooks in your IDE)
+* [ ] Check our [Contribution Guidelines](https://github.com/corona-warn-app/cwa-server/blob/master/CONTRIBUTING.md)
```
johanneseschrig, line 14: I think these points are meta points on opening a PR, and should be in the commented out section above. The actual checklist should only contain checks relevant to the functionality, something like: 

- [ ] Test your changes as thoroughly as possible before you commit them. Preferably, automate your test by unit/integration tests.
- [ ] Make sure `mvn install` runs for the whole project and both submission and distribution service can be run with `spring-boot:run`

PR #373, opened 2020-05-29
Update file headers with maven plugin generated license headers

stevesap:
As per [license-maven-plugin](https://www.mojohaus.org/license-maven-plugin/) integrated in https://github.com/corona-warn-app/cwa-server/pull/362

- Updated file headers with maven plugin generated license headers
- Added missing headers 
- Excluding any classes under io.minio.* as they require special license


CCFenner:
please mind #374 regarding the checkstyle issues

PR #376, opened 2020-05-29
Fixes wait-for-it.sh docker script execution (closes #339)

### LINKED ISSUE #339: Fix docker compose wait-for-it.sh script
### Current status

The [execution ](https://github.com/corona-warn-app/cwa-server/blob/7d1a5a4e293a83a388d126f32a1c229f13467437/docker-compose.yml#L87)of the existing [wait-for-it.sh script](https://github.com/corona-warn-app/cwa-server/blob/master/scripts/wait-for-it/wait-for-it.sh) was reverted in https://github.com/corona-warn-app/cwa-server/pull/338 since it was not executing properly in windows.

Error message from ``docker-compose run create-bucket`` is:

``No such file or directory``

### Expected Behavior

wait-for-it.sh should execute properly in linux, mac and windows environments

stevesap:
Issue was with execution of script on windows.

- LF EOLs are required for proper execution - enforcing LF EOL via `.gitattributes` file so that windows git setups do not covert to CRLF on checkout and the file always gets commited to repo with LF, regardless of client
- Addition of `aws` command for `s3api` required for proper execution
- `wait-for-it.sh` file needed touching in order for EOL conversion to take effect in git repo

NOTE: WIP as PR is still pending testing on linux setup.


AlexanderBarbie:
Works on MacOS.

Great solution, didn't know you can set the EOL of single files in gitattributes. :)

PR #378, opened 2020-05-29
Re-introduce AWS SDK dependency

### LINKED ISSUE #363: Remove MinIO SDK and use AWS SDK instead
Since MinIO SDK still hasn't released a fix version for fixing the XML vulnerability in the SimpleXML library, we need to consider alternative S3 SDK's. Therefore, prepare an alternative approach with AWS SDK, which is then used by the ObjectStoreAccess class.

michael-burwig:
Resolves #363 

- Replaces minio dependency with AWS sdk - s3 dependency.
- Replaces code that relies on ``google.commons`` dependency (transitive dependency of minio).
- Introduces ``S3ClientWrapper`` (implements ``ObjectStoreClient``).
- Implements dummy ``AwsCredentialsProvider`` that serves credentials based on the respective ``DistributionServiceConfig`` fields.
- Removes ports from default object store ``endpoint`` values in ``apllication.yaml``.


services/distribution/src/main/java/app/coronawarn/server/services/distribution/objectstore/client/S3ClientWrapper.java
```diff
@@ -0,0 +1,122 @@
+/*
```
stevesap, line 1: Replace with new license file header for consistency

services/distribution/src/test/java/app/coronawarn/server/services/distribution/objectstore/client/S3ClientWrapperTest.java
```diff
@@ -0,0 +1,237 @@
+/*
```
stevesap, line 1: Replace with new license file header for consistency

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/diagnosiskeys/structure/file/TemporaryExposureKeyExportFile.java
```diff
@@ -132,7 +138,12 @@ public void prepare(ImmutableStack<Object> indices) {
   private byte[] getHeaderBytes() {
     String header = distributionServiceConfig.getTekExport().getFileHeader();
     int headerWidth = distributionServiceConfig.getTekExport().getFileHeaderWidth();
-    return Strings.padEnd(header, headerWidth, ' ').getBytes(StandardCharsets.UTF_8);
+    return padRight(header, headerWidth).getBytes(StandardCharsets.UTF_8);
+  }
+
+  private String padRight(String string, int padding) {
+    String format = "%1$-" + padding + "s";
```
johanneseschrig, line 145: why the extra assignment?

PR #382, opened 2020-05-29
Set maximum total risk score to 255

### LINKED ISSUE #381: Wrong maximum total risk score in risk score classifications
## Describe the bug

Currently, the maximum value for a total risk score class is set to 4096.

## Expected behaviour

The [Apple documentation](https://developer.apple.com/documentation/exposurenotification/enexposureconfiguration) states:
> Level values are in the range of 0-8. Though the formula’s range can be up to 4096, the framework limits totalRiskScore to the upper limit of ENRiskScore, which has a maxium value of 255.

According to @jakobmoellersap, while Google does not seem to state this explicitly, this is reflected in their implementation.

michael-burwig:
Fixes #381 

- Adjusts the maximum total risk score.
- Adjusts tests.
- Adjusts master-config.

### DIFF
--- a/services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/appconfig/validation/RiskScoreClassificationValidator.java
+++ b/services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/appconfig/validation/RiskScoreClassificationValidator.java
@@ -81,7 +81,7 @@ public class RiskScoreClassificationValidator extends ConfigurationValidator {
   }
 
   private void validateRiskScoreValueBounds(int value) {
-    if (!RiskScoreValidator.isInBounds(value)) {
+    if (value < 0 || value > ParameterSpec.TOTAL_RISK_SCORE_MAX) {
       errors.add(new RiskScoreClassificationValidationError("minRiskLevel/maxRiskLevel", value, VALUE_OUT_OF_BOUNDS));
     }
   }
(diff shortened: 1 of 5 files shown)


services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/appconfig/validation/RiskScoreClassificationValidator.java
```diff
@@ -81,7 +81,7 @@ private void validateLabel(String label) {
   }
 
   private void validateRiskScoreValueBounds(int value) {
-    if (!RiskScoreValidator.isInBounds(value)) {
+    if (value < 0 || value > ParameterSpec.TOTAL_RISK_SCORE_MAX) {
```
stevesap, line 84: Consider introducing `ParameterSpec.TOTAL_RISK_SCORE_MIN` for consistency with `ParameterSpec.TOTAL_RISK_SCORE_MAX` and `ParameterSpec.RISK_SCORE_MIN`

christian-kirschnick:
We are missing out the min-risk-score in the app-config.yaml. It is also bound to the same datatype, so it should be included. This would also mean, that you would just need to change the parameter spec max value. However, still waiting for the mobile team to give further input on this, as there are some open questions.

christian-kirschnick:
Talked to mobile team - they will investigate further. In the meantime, do not continue or merge this PR.

christian-kirschnick:
Was a bug on their end, so I'll close this one.

PR #387, opened 2020-05-30
Improve external tan verification

### LINKED ISSUE #327: Use typed client and entities for tan verification
## Current Implementation
The current tan verification look like this:
https://github.com/corona-warn-app/cwa-server/blob/3b98beb268bac511c00f8e13bcbb4978a8cdb4fd/services/submission/src/main/java/app/coronawarn/server/services/submission/verification/TanVerifier.java#L102-L115

There are a couple of issues I would see here:
1. The tan entity is build from a string which can cause bugs, encoding issues an more.
2. When the interface of the verification server changes (and this a repo owned by the same organization) this has to be modified as well and can cause runtime bugs without proper system testing

## Suggested Enhancement
I would suggest two things
1. Create a DTO (Pojo) for the tan payload and pass that to the rest template to let Jackson handle serialization
2. If a dependency between cwa projects is "allowed" (which I would think is perfectly fine)  I would actually move that entity to the verification server project in a separate maven module. I would add [Spring Feign client ](https://cloud.spring.io/spring-cloud-netflix/multi/multi_spring-cloud-feign.html) and use that in the CWA server. Feign is like the reverse version of `@RestController` and supports the same annotations. It will create the client implementation from the given interface at runtime.

The client would look like this:

```
@FeignClient(name = "verification-server", url = "${services.submission.verification.baseUrl}")
public interface StoreClient {

    @RequestMapping(method = RequestMethod.POST, value = "${services.submission.verification.path}", consumes = "application/json")
    ValidationResult validateTan(Tan tan); // both classes should be Pojos
}
```

The usage would look like this (I did not actually implemented that!):

```
 private boolean verifyWithVerificationService(String tanString) {
   Tan tan = Tan.from(tanString); // This should also do the UUID verification!
    
    ValidationResult result = verificationServer.validateTan(tan);

    return result != null;
  }
```

## Expected Benefits
- Compilation breaks when interface changes
- Less risk of bugs

jwedel:
This PR introduces a domain object for TANs that also does input validation. Moreover, a spring cloud feign based client is introduced to call the external HTTP service using that new entity. The existing tests have been split to Tan class tests for syntax checking and using a SpringBoot integration test with WireMock to actually test the feign client with real HTTP communication.

Resolves #327


services/submission/src/test/java/app/coronawarn/server/services/submission/verification/FeignTestConfiguraiton.java
```diff
@@ -0,0 +1,15 @@
+package app.coronawarn.server.services.submission.verification;
+
+import org.springframework.boot.autoconfigure.http.HttpMessageConverters;
+import org.springframework.cloud.openfeign.EnableFeignClients;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@EnableFeignClients
+@Configuration
+public class FeignTestConfiguraiton {
```
christian-kirschnick, line 10: Typo

services/submission/src/main/java/app/coronawarn/server/services/submission/verification/VerificationServerClient.java
```diff
@@ -0,0 +1,10 @@
+package app.coronawarn.server.services.submission.verification;
+
+import org.springframework.cloud.openfeign.FeignClient;
+import org.springframework.web.bind.annotation.PostMapping;
+
+@FeignClient(name = "verification-server", url = "${services.submission.verification.base-url}")
+public interface VerificationServerClient {
```
christian-kirschnick, line 7: Add some Java Doc here, same goes for the verifyTan method.

services/submission/src/main/java/app/coronawarn/server/services/submission/verification/VerificationServerClient.java
```diff
@@ -0,0 +1,10 @@
...
+  @PostMapping(value = "${services.submission.verification.path}", consumes = "application/json")
+  String verfiyTan(Tan tan);
```
christian-kirschnick, line 9: typo - also would be nice to use `MediaType.APPLICATION_JSON` here instead of the string literal

jwedel:
@christian-kirschnick 
Corrected the typos and Checkstyle issues. That's why a added the tab size to .editorconfig, but I did this after creating a couple of files already.

On circle-ci, I see some runtime errors I don't have locally. I need to investigate that.

However, locally, I get errors in the distribution test suite:

```
Caused by: java.lang.ClassCastException: class org.bouncycastle.asn1.pkcs.PrivateKeyInfo cannot be cast to class org.bouncycastle.openssl.PEMKeyPair (org.bouncycastle.asn1.pkcs.PrivateKeyInfo and org.bouncycastle.openssl.PEMKeyPair are in unnamed module of loader 'app')
 ```

I'm building with OpenJDK 14. It seems that some crypto-provider-related issues there. Have you successfully built it with 14?

File: .editorconfig [hunk 4]
```diff
@@ -2,6 +2,7 @@ root = true
 
 [*]
 indent_style = space
+indent_size = 2
```
michael-burwig, line 5: Unfortunately, this is messing up the consistency with the existing code base in intellij. E.g. wrapped method signatures currently are indented with 4 spaces.

jwedel: Ok, but what’s the point on using editorconfig when the editor should not use it.

I'm using IntelliJ and with the given config, all indentation is set to 4 b default.

In one of the more recent versions Intellj, more detailed settings for editorconfig is supported (https://www.jetbrains.com/help/idea/configuring-code-style.html).

I can revert it, but how can I format it automatically then?

michael-burwig: Hi Jan,
sorry that this review is taking a little longer - busy days.
I will propose adding intelliJ specific entries to the editor config either later tonight or tomorrow.

jwedel: No problem, I really appreciate all your feedback. You're all doing a great job here, working on the weekends and late each day.

services/submission/pom.xml
```diff
@@ -48,6 +48,18 @@
       <groupId>io.micrometer</groupId>
       <artifactId>micrometer-registry-prometheus</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.springframework.security</groupId>
+      <artifactId>spring-security-test</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.github.tomakehurst</groupId>
```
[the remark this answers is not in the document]
jwedel: I've added WireMock and Spring Cloud. Spring security test just has been moved to the bottom of the list because it's test scope.

.editorconfig
jwedel: By the way, if you have a recent version of IntelliJ and a working ImtelliJ code style configuration, you can just create a new editorconfig file (New -> editorconfig) and it will
generate all the necessary entries.

Generally speaking, the easiest would be to just go with the language default settings and don’t customize. But I guess that’s not an option...

christian-kirschnick:
FYI: I pushed your branch of the fork to the repo, so Sonar can also have a look:
https://sonarcloud.io/dashboard?branch=fix%2Fimprove-tan-verifier&id=corona-warn-app_cwa-server

jwedel:
Thx, I will have a look into it hopefully this evening.

Issue #354, opened 2020-05-28
Use license-maven-plugin to ensure file headers and third party license file generation


johanneseschrig:
## Feature description

We should consider the use of [license-maven-plugin](https://www.mojohaus.org/license-maven-plugin/)

## Problem and motivation

Currently, adding a license header is a manual task, as well as updating the third party license file. This should optimally be automated.

stevesap:
[license-maven-plugin](https://www.mojohaus.org/license-maven-plugin/) was integrated in https://github.com/corona-warn-app/cwa-server/pull/362 for updating file license headers via manual command `mvn license:update-file-header`

Remaining tasks on this issue:
1. Generate the [THIRD-PARTY-LICENSE](https://github.com/corona-warn-app/cwa-server/blob/master/THIRD-PARTY-NOTICES) file
2. Automate the maven plugin execution

CCFenner:
I was able to generate a `THIRD-PARTY-NOTICE` file with the `license-maven-plugin`. However this only contains the dependencies (maven-GAV, url, name) and not the actual license text. Would this already be sufficient?

stevesap:
@tkowark can you please advise if such a generated `THIRD-PARTY-LICENSE` file as @CCFenner described would be sufficient?

stevesap:
@CCFenner how about a blocking Github Action which verifies that every relevant file in PR has the correct license header? If not, the PR author can then correct the issues locally either manually or by using the `license-maven-plugin` (to be documented in CONTRIBUTION.md)

As for the THIRD-PARTY-LICENSE file, pending on @tkowark 's confirmation, we could try a similar approach with respect to validating a correct THIRD-PARTY-LICENSE file as a result of relevant PR changes.

CCFenner:
> @CCFenner how about a blocking Github Action which verifies that every relevant file in PR has the correct license header? If not, the PR author can then correct the issues locally either manually or by using the `license-maven-plugin` (to be documented in CONTRIBUTION.md)

Should be doable, I'll have a look at it tonight. Would it be sufficient to have it blocking on any missing header not only changed files? I guess this would make the job much simpler.

stevesap:
@CCFenner thanks. Yes I think that would be sufficient if it makes it simpler.

tkowark:
@CCFenner @stevesap Sorry for the delay in response. Please automate that process of collecting all dependencies and their licenses. For the time being, let's collect the license texts manually (they should not change that often). 

[...]

PR #388, opened 2020-05-30
Ensure uniqueness of diagnosis keys

### LINKED ISSUE #289: Ensure uniqueness of diagnosis keys
## Current Implementation
The uniqueness of diagnosis keys is neither guaranteed by the database, nor by validation logic.

## Suggested Enhancement
Ensure that diagnosis keys in the database are unique (there are no duplicates).

What needs to be decided:
- [x] Is it sufficient to guarantee the uniqueness of the diagnosis key or should the uniqueness of diagnosis key + submission timestamp (+maybe even more fields) be guaranteed?
- [x] Should the duplicate check happen on the database (`UNIQUE` constraint) or in the server logic (e.g. during the validation of submitted keys)?
- [x] If the check happens on the database: Should the `id` column be removed and replaced with the diagnosis key (+submission timestamp) as (composite) primary key?
- [x] How should the server react when a duplicate diagnosis key is submitted (update, error, etc.)?

**Edit:** Decisions are below.

UnchartedBull:
This PR removes the ID field from the diagnosis keys table and places the actual key as the primary key instead. The postgres database schema has been updated as well.

Hibernate unfortunately does not support `ON CONFLICT`, thus I had to overwrite the saveAll method from the DiagnosisKeyRepository to make sure, that we ignore the new key. Default behaviour from Hibernate is to update the existing entry, which can't be changed as far as I know.

Closes #289.


common/persistence/src/test/java/app/coronawarn/server/common/persistence/service/DiagnosisKeyServiceTest.java
```diff
@@ -149,15 +150,39 @@ void testNoPersistOnValidationError() {
...
+
+    var actKeys = diagnosisKeyService.getDiagnosisKeys();
+
+    assertThat(actKeys.size()).isEqualTo(1);
+    assertThat(actKeys.iterator().next().getTransmissionRiskLevel()).isEqualTo(2);
+  }
+
+  public DiagnosisKey buildDiagnosisKeyForSubmissionTimestamp(long submissionTimeStamp) {
+    byte[] randomBytes = new byte[16];
+    Random random = new Random(submissionTimeStamp);
```
johanneseschrig, line 176: With regards to reproducible test results, should we consider not generating random input values for tests?

common/persistence/src/main/resources/db/migration/postgres/V1__createTables.sql
```diff
@@ -1,6 +1,5 @@
 CREATE TABLE diagnosis_key (
-    id bigserial PRIMARY KEY,
-    key_data bytea,
+    key_data bytea PRIMARY KEY,
```
johanneseschrig, line 2: This needs to be an own V3 migration, else this will crash on existing systems. (Use `ALTER TABLE`)

common/persistence/src/main/java/app/coronawarn/server/common/persistence/repository/CustomDiagnosisKeyRepository.java
```diff
@@ -0,0 +1,26 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package app.coronawarn.server.common.persistence.repository;
+
+import java.util.List;
+
+public interface CustomDiagnosisKeyRepository<T> {
```
johanneseschrig, line 24: Consider not using interfaces if only a single implementation exists. Or are further `CustomDiagnosisKeyRepository`s planned? Then the implementation should not carry the `Impl` suffix, but be more descriptive.

common/persistence/src/main/java/app/coronawarn/server/common/persistence/repository/CustomDiagnosisKeyRepositoryImpl.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.transaction.annotation.Transactional;
+
+@NamedQuery(
+    name = "findAllCustomersWithName",
+    query = "INSERT INTO diagnosis_key VALUES {:custName,144,2,441907,1};"
+)
+
+@Transactional
+public class CustomDiagnosisKeyRepositoryImpl implements CustomDiagnosisKeyRepository<DiagnosisKey> {
```
johanneseschrig, line 35: Consider not using the `Impl` suffix.

common/persistence/src/main/java/app/coronawarn/server/common/persistence/repository/CustomDiagnosisKeyRepositoryImpl.java
```diff
@@ -0,0 +1,49 @@
...
+
+import app.coronawarn.server.common.persistence.domain.DiagnosisKey;
+import java.util.List;
+import javax.persistence.EntityManager;
+import javax.persistence.NamedQuery;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.transaction.annotation.Transactional;
+
+@NamedQuery(
+    name = "findAllCustomersWithName",
```
johanneseschrig, line 30: customers?

common/persistence/src/test/java/app/coronawarn/server/common/persistence/service/DiagnosisKeyServiceTest.java
```diff
@@ -149,15 +150,39 @@ void testNoPersistOnValidationError() {
...
     return DiagnosisKey.builder()
-        .withKeyData(new byte[16])
+        .withKeyData(randomBytes)
         .withRollingStartIntervalNumber(600)
         .withTransmissionRiskLevel(2)
         .withSubmissionTimestamp(submissionTimeStamp).build();
   }
 
-  public static DiagnosisKey buildDiagnosisKeyForDateTime(OffsetDateTime dateTime) {
+  public DiagnosisKey buildDiagnosisKeyForDateTime(OffsetDateTime dateTime) {
```
michael-burwig, line 185: why not ``static``?

common/persistence/src/main/java/app/coronawarn/server/common/persistence/repository/CustomDiagnosisKeyRepository.java
```diff
@@ -0,0 +1,26 @@
+/*
```
stevesap, line 1: Please replace with with generated license file header by copying from other files or just running `mvn license:update-file-header` on root. (This needs to be documented still but pending https://github.com/corona-warn-app/cwa-server/issues/354)

common/persistence/src/main/java/app/coronawarn/server/common/persistence/repository/CustomDiagnosisKeyRepositoryImpl.java
```diff
@@ -0,0 +1,49 @@
+/*
```
stevesap, line 1: Please replace with with generated license file header by copying from other files or just running `mvn license:update-file-header` on root. (This needs to be documented still but pending https://github.com/corona-warn-app/cwa-server/issues/354)

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/DiagnosisKey.java
```diff
@@ -45,6 +45,12 @@
  */
 @Entity
 @Table(name = "diagnosis_key")
+@NamedNativeQuery(
+    name = "DiagnosisKey.saveAllDoNothingOnConflict",
+    query = "INSERT INTO diagnosis_key"
```
johanneseschrig, line 50: Just a note: Make sure to test on H2, also.

common/persistence/src/test/java/app/coronawarn/server/common/persistence/service/DiagnosisKeyServiceTest.java
UnchartedBull: We need different diagnosis keys here, since the second one wouldn't be saved otherwise, due to the new constraints on the db. This whole random thing, isn't really random. Having the same random seed (submission timestamp in this case) will result in the same bytes being generated.

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/DiagnosisKey.java
michael-burwig: ✅

PR #394, opened 2020-05-31
Use native AWS sdk credentials provider

michael-burwig:


services/distribution/src/main/java/app/coronawarn/server/services/distribution/objectstore/client/ObjectStoreClientConfig.java
```diff
@@ -36,47 +37,26 @@
...
-          return accessKeyId;
-        }
-
-        @Override
-        public String secretAccessKey() {
-          return secretAccessKey;
-        }
-      };
-    }
+  private String removeTralingSlash(String string) {
```
johanneseschrig, line 59: tpyo

johanneseschrig: image

k

PR #397, opened 2020-05-31
Add Google-Style IntelliJ formatter settings to .editor.conf

michael-burwig:


.editorconfig
```diff
@@ -1,8 +1,253 @@
 root = true
 
 [*]
-indent_style = space
-end_of_line = lf
 charset = utf-8
+end_of_line = lf
```
stevesap, line 5: FYI - On Windows, files checked out by git start out with CRLF line ending by default, but then due to this setting, get converted to LF (at least by my Windows InteliJ IDE) upon modifying and saving a file. However, this doesn't seem to cause any trouble as Windows InteliJ is probably smart enough to handle LF fine as well. And as we've already had this (end_of_line) in for a while and some of us are Windows users, this is probably safe to keep.

.editorconfig
```diff
@@ -1,8 +1,253 @@
...
+indent_size = 2
+indent_style = space
 trim_trailing_whitespace = true
 insert_final_newline = true
+max_line_length = 100
```
stevesap, line 10: @michael-burwig can you confirm we want max_line_length here as 100 and not 120?

.editorconfig
Bananeweizen: @stevesap You may eventually want to configure core.autocrlf for that repository to make git enforce a certain line ending style: https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration (in case you don't trust the IDE or use other editors additionally).

PR #407, opened 2020-06-01
Calculate real-time submission delay with CAS instead of synchronization

### LINKED ISSUE #398: Unnecessary synchronized
https://github.com/corona-warn-app/cwa-server/blob/822150b1b22645ba071a6f6576d50614dc01b34b/services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java#L168

Synchronization at this point causes frequent locks which will harm performance. It is not required if a local final value is used instead:

```
private void updateFakeDelay(final long realRequestDuration) {
  final Double currentDelay = fakeDelay;
  fakeDelay = currentDelay + (1 / fakeDelayMovingAverageSamples) * (realRequestDuration - currentDelay);
}
```

`fakeDelay` could be marked volatile if up-to-date values are required, but as far as I can tell from the code this isn't necessary.

arteam:
## Checklist

## Description

Instead of synchronizing on every real submission request, we can use a CAS to update the current submission delay based on optimistic locking. Using the CAS, threads spin lock in the user-mode instead of waiting and getting woken up by the OS.

Fixes #398


services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java
```diff
@@ -164,7 +166,18 @@ public void persistDiagnosisKeysPayload(SubmissionPayload protoBufDiagnosisKeys)
...
 
-  private synchronized void updateFakeDelay(long realRequestDuration) {
-    fakeDelay = fakeDelay + (1 / fakeDelayMovingAverageSamples) * (realRequestDuration - fakeDelay);
+  /** @implNote VisibleForTesting */
+  void updateFakeDelay(long realRequestDuration) {
+    while (true) {
+      Double fd = fakeDelay.get();
+      if (fakeDelay.compareAndSet(
+          fd, fd + (1 / fakeDelayMovingAverageSamples) * (realRequestDuration - fd))) {
+        break;
```
twothe, line 175: `while (true) .. break` is not a good code style.

For the code it seems irrelevant which thread changes fakeDelay at which point (something this code does not fix either), therefore compareAndSet is unnecessary. The only relevant thing is to make sure that the first fakeDelay value does not differ from the second fakeDelay value in the above calculate. And that a simple final value would do, using an Atomic in this case is overkill.

stevesap:
Thanks for your efforts @arteam. However, we decided to go with the solution in https://github.com/corona-warn-app/cwa-server/pull/455 for the reasons discussed there.

PR #409, opened 2020-06-01
Fixing S3ClientWrapper's bucketExists method

yvz5:
## Description

The old implementation queried the whole bucket list and checked only the first item whithout considering the bucket name. So the bug is that the searched bucketname would never be used.

In a newer aws sdk there is already a method checking for existing buckets. Consider upgrading from [S3Client ](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/AmazonS3Client.html) to [AmazonS3](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/index.html?com/amazonaws/services/s3/model/Bucket.html).

This code change uses [HeadBucketRequest ](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/index.html?com/amazonaws/services/s3/model/Bucket.html) object to query using the wanted bucket name which moves the search logic to the aws side, thus better performance & less costs.


stevesap:
Hi @yvz5 thanks for your very first contribution to our project! Coincidentally, we also noticed the issue at the same time and were working to resolve it. 

We would however like to change the implementation a bit since `headBucket` call throws exception and so the following return will not be evaluated if bucket does not exist. Also, we are trying to limit the required permissions (`headBucket` requires `S3:listbucket`). You can see the changes we'd like to make [in this branch](https://github.com/corona-warn-app/cwa-server/commit/da7a0560a2eebe57c876a5413da1577a683a1ee2).

Are you OK with us pushing the changes directly to your PR or do you prefer we create a new PR and subsequently close yours? A prompt response would be greatly appreciated! 👍

yvz5:
> Hi @yvz5 thanks for your very first contribution to our project! Coincidentally, we also noticed the issue at the same time and were working to resolve it.
> 
> We would however like to change the implementation a bit since `headBucket` can throw `NoSuchBucket` Exception and so the following return will not be evaluated if bucket does not exist. Also, we are trying to limit the required permissions (`headBucket` requires `S3:listbucket`). You can see the changes we'd like to make [in this branch](https://github.com/corona-warn-app/cwa-server/commit/da7a0560a2eebe57c876a5413da1577a683a1ee2).
> 
> Are you OK with us pushing the changes directly to your PR or do you prefer we create a new PR and subsequently close yours? A prompt response would be greatly appreciated! 👍

Hi steve, 
I would prefer using the builtin method doesBucketExistV2 but I assume that would require a great deal of changes. 

I would actually be glad if you commit to this PR.

michael-burwig:
Thanks for pointing out this defect.

As far as I can tell ``AmazonS3.doesBucketExistV2`` is available in the older AWS SDK version 1.X only.

PR #421, opened 2020-06-02
add workflow to ensure license file headers (#354)

CCFenner:
## Description

This change adds an action to verify the license file headers in `*.java` files using the [license-maven-plugin](https://www.mojohaus.org/license-maven-plugin/). See [here](https://github.com/CCFenner/cwa-server/pull/6/checks?check_run_id=731696193) how it looks when the check fails.

related to https://github.com/corona-warn-app/cwa-server/issues/354#issuecomment-637369904


CCFenner:
I created this as GH action because I'm used to it. This could of course be added to the CircleCI workflow as well.

PR #424, opened 2020-06-02
Feature/app version config

### LINKED ISSUE #416: Application Configuration: Add version update support
We would like to keep the mobile application informed about new releases/current release of the app. This information can then be presented to the user, in order to ask for applying updates. 

We would like to add new fields to the application config (example values):
 
```yaml
app-version:
    ios:
        latest: 1.5.2
        min: 1.2.2
    android: 
        latest: 1.3.2
        min: 1.1.2
```

After downloading the config, the app checks whether the signature can be verified. If that is the case, it will get the current & min values from the config based on the target platform. 

This information can now be used to prompt users for updating the app. The values for the versions must follow semantic versioning.

pithumke:
Closes #416

Summary:
- Add "app-version" configuration


services/distribution/src/main/resources/master-config/app-version-config.yaml
```diff
@@ -0,0 +1,25 @@
+# This is the Application Version Configuration master file which contains information for clients
+# about the latest and minimum supported mobile app versions on Android and iOS.
+#
+# The latest version mus not be lower than the min version.
```
johanneseschrig, line 4: typo

PR #450, opened 2020-06-04
Logging: Add logger statements for Verification Server calls (closes #430)

### LINKED ISSUE #430: Logging: Add logger statements for Verification Server calls
When the verification server is called, we should print logger statements before communication starts and after the communications ends, which allows us to further pinpoint potential problems with the verification server.

I checked this with DPP and we got their OK.

stevesap:
## Checklist

## Description

Added logging as indicated in linked ticket.


stevesap:
@MKusber please review log messages

MKusber:
Please consider making the last message an info instead.

christian-kirschnick:
Actually all those logger statements should be INFO.

PR #459, opened 2020-06-05
Add mutual TLS

pithumke:


services/submission/src/main/java/app/coronawarn/server/services/submission/verification/CloudFeignClientProvider.java
```diff
@@ -0,0 +1,70 @@
...
+import org.springframework.context.annotation.Profile;
+import org.springframework.core.env.Environment;
+import org.springframework.stereotype.Component;
+import org.springframework.util.ResourceUtils;
+
+@Component
+@Profile("ssl-client-verification")
+public class CloudFeignClientProvider implements FeignClientProvider {
+
+  Environment environment;
```
christian-kirschnick, line 39: @pithumke @MKusber fyi Changed to constructor autowiring for consistency.

PR #460, opened 2020-06-05
Switch to CWA-Hash from e-Tag

christian-kirschnick:


services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/structure/archive/ArchiveOnDisk.java
```diff
@@ -102,4 +107,19 @@ public void prepare(ImmutableStack<Object> indices) {
   public void setBytes(byte[] bytes) {
     throw new UnsupportedOperationException("Can not set bytes on an archive.");
   }
+
+  private void updateChecksumBytes() {
+    var targetFile = this.getWritables().stream()
+      .filter(writable -> writable instanceof File)
```
michael-burwig, line 113:
```suggestion
        .filter(Writable::isFile)
```

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/structure/file/FileOnDisk.java
```diff
@@ -64,4 +68,8 @@ public void setBytes(byte[] bytes) {
   public void prepare(ImmutableStack<Object> indices) {
     // Method override exists here to comply with the implementation rules for the Writable interface.
   }
+
+  public byte[] getBytesForChecksum() {
```
michael-burwig, line 72: never called

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/structure/file/FileOnDiskWithChecksum.java
```diff
@@ -0,0 +1,101 @@
...
+    IO.makeNewFile(checksumFile.getParent().toFile(), checksumFile.getFileName().toString());
+    IO.writeBytesToFile(checksum, checksumFile.toFile());
+  }
+
+  /**
+   * Fetches the target bytes for computation of the checksum. Will take the bytes of its {@link FileOnDisk}.
+   *
+   * @return the checksum bytes
+   */
+  public byte[] getBytesForChecksum() {
```
michael-burwig, line 71: can be ``private``

services/distribution/src/main/java/app/coronawarn/server/services/distribution/objectstore/ObjectStoreAccess.java
```diff
@@ -32,6 +32,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
+import org.springframework.util.StringUtils;
```
michael-burwig, line 35: From the doc:
```
 * <p>Mainly for internal use within the framework; consider
 * <a href="https://commons.apache.org/proper/commons-lang/">Apache's Commons Lang</a>
 * for a more comprehensive suite of {@code String} utilities.
```

In our case we could rely on ``String.isEmpty`` since it seems guaranteed that ``getChecksum`` returns at least ``""``.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/structure/file/FileOnDisk.java
christian-kirschnick: It is - and it is also overwritte in ArchiveOnDisk. So lets make it protected.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/objectstore/ObjectStoreAccess.java
christian-kirschnick: Still like the "safer" version here.

johanneseschrig:
Please fix checkstyle.

PR #466, opened 2020-06-06
S3 compatible storage multithreading

### LINKED ISSUE #150: S3Publisher: Increase performance
The S3Publisher is currently running in only one thread, making it quite slow for uploads/requests for metadata.

Enhance the S3Publisher/ObjectStoreAccess, so that multiple threads are being used for interacting with S3.

UnchartedBull:
Enables the s3 put object operation to be run in multiple threads. The number of threads can be configured in the distribution service configuration.

Closes #150.


services/distribution/src/main/java/app/coronawarn/server/services/distribution/objectstore/ObjectStoreAccess.java
```diff
@@ -86,8 +89,10 @@
    *
    * @param localFile The file to be published.
    */
-  public void putObject(LocalFile localFile) {
+  @Async("s3TaskExecutor")
+  public Future<Void> putObject(LocalFile localFile) {
     putObject(localFile, DEFAULT_MAX_CACHE_AGE);
+    return new AsyncResult<>(null);
```
christian-kirschnick, line 95: looks weird - maybe the result of putObject should also be returned here

services/distribution/src/main/java/app/coronawarn/server/services/distribution/util/AsyncConfiguration.java
```diff
@@ -0,0 +1,64 @@
...
+  /**
+   * Creates an Executor, which is used by {@link S3Publisher} to multi-thread the S3 put operation. Requests to {@link
+   * ObjectStoreAccess} putObject method will be automatically proxied by SpringBoot and thus run multithreaded.
+   * Daemonized threads need to be used here, in order to allow Java to "naturally" terminate once all threads are
+   * finished.
+   *
+   * @return the executor, which tells SpringBoot the basic parameters.
+   */
+  @Bean(name = "s3TaskExecutor")
+  public Executor taskExecutor() {
```
christian-kirschnick, line 56: Could rename the method to s3TaskExecutor, so the bean name is not necessary anymore

services/distribution/src/main/java/app/coronawarn/server/services/distribution/util/DaemonThreadFactory.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ---license-end
+ */
+
+package app.coronawarn.server.services.distribution.util;
+
+import java.util.concurrent.ThreadFactory;
+
+public class DaemonThreadFactory implements ThreadFactory {
```
christian-kirschnick, line 25: Rename so it's clear that this thread factory is for our S3 upload workers

services/distribution/src/main/java/app/coronawarn/server/services/distribution/objectstore/ObjectStoreAccess.java
UnchartedBull: s3client.putObject is a void method, so there is nothing to return here. We still need the Future to determine, whether the thread has finished or not.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/util/S3UploadDaemonThreadFactory.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ---license-end
+ */
+
+package app.coronawarn.server.services.distribution.util;
```
michael-burwig, line 21: Stolen from the internet:

> ### Util Packages Are Evil
> util, common, base, infrastructure, model, framework names sounds great but are too generic as package names. Such names are un-cohesive by nature. Because of the abstract naming, they end up being a "kitchen sink". All inconvenient classes are placed in the "util" packages what makes such packages bloated, the classes harder to find and the application harder to maintain.
Instead of naming a package util, name it after it true responsibilities e.g. logging, monitoring etc. It will keep the packages small and the application easier to maintain.

michael-burwig:
- upload operation threadpool size set to 4
- added ``@Component`` to``S3Publisher``

PR #467, opened 2020-06-06
[trivial] Fix postgres endpoint in README

jsievers:
postgres hostname should be 'localhost'


jsievers:
I was trying out "Running the Full CWA Backend" from the README and while playing around with the various endpoints I noticed a trivial typo for the postgres hostname.

PR #478, opened 2020-06-08
Number of bytes should not depend on Locale

jsievers:
Charset.defaultCharset() depends on the locale
of the JVM and OS. Tests would fail e.g. with
UTF-16 (2 byte per character) encoding. Use US-ASCII
instead to make it obvious we use a 1 byte per
character encoding for test data.


jsievers:
I think there are more places in the codebase where java.lang.String.getBytes() is used which is the same problem but not sure  if encoding is really relevant in all cases. See e.g. https://github.com/corona-warn-app/cwa-server/blob/d9e5e0156f322ca386066d85f8aecd39a75f243d/services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/structure/directory/decorator/indexing/IndexingDecoratorOnDisk.java#L49

jsievers:
thanks! quick turnaround.

PR #495, opened 2020-06-09
remove use of ForkJoinPool in SubmissionController (closes #399)

### LINKED ISSUE #399: Dangerous use of ForkJoinPool
https://github.com/corona-warn-app/cwa-server/blob/822150b1b22645ba071a6f6576d50614dc01b34b/services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java#L63

The common ForkJoinPool is used to process all real incoming transmissions. This will cause the pool to be flooded under heavy load and block all other threaded streaming operations, which could potentially put the whole application into a dead-lock. 

Also no forking and joining is done in the processing, which means the wrong type is used here. You should to use a standard Executor instead.

In general the DeferredResult should only be used if there is no alternative, as it uses dangerous locking, unnecessary synchronization and the [double-checked locking anti-pattern](https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java).

nils4cosee:
As discussed in #399:

While a scheduled-executor makes sense for timing the response to
fake-requests, real requests should better be handled in the web-thread
directly. We still use the DeferredResult to as response, but do not
run real requests in new threads.

## Checklist


nils4cosee:
@michael-burwig We noticed that you cannot switch on "maintainers can commit" in a fork of an organisation, sorry about that.

michael-burwig:
My apologies again for not having picked this PR after all. The issue in question became more and more pressing due to performance/high availability tests that were scheduled on that day.

PR #523, opened 2020-06-10
RC Patch 1.0.2

ole-lilienthal:


ChristopherSchmitz:
@pithumke - please not deploy this change to WRU. If deployed we have to revert the change. Since android: 0.9.0 is to high we will trigger the "force-update" on all testers which prevent them to work on the test cases
@HeeTattSap

pithumke:
Ok, [reverted](https://github.com/corona-warn-app/cwa-server/pull/524) and [corrected](https://github.com/corona-warn-app/cwa-server/pull/529).

PR #525, opened 2020-06-11
Add TLS Ciphers compliant to BSI-TR-02102-2

MKusber:
This will update the offered cipher suites to comply with mentioned technical guidance by BSI.
Still, this supported domain parameters need to be adapted.


MKusber:
Please do not merge yet. Will have a short review with @christian-kirschnick to clarify whether this is a sufficient solution.

PR #529, opened 2020-06-11
Update app versions

pithumke:
```
ios.latest: 0.8.2
ios.min: 0.5.0
android.latest: 0.8.3
android.min: 0.8.0
```

@ChristopherSchmitz


ChristopherSchmitz:
ok - thanks for your support

PR #533, opened 2020-06-11
Replace H2 with Docker based testcontainers-postgres for testing

kiview:
Instead of using an H2 in-memory database for testing, a real PostgreSQL database is launched on-demand as a Docker container.

## Description

This PR allows integration tests to be executed against a real PostgreSQL database, thereby giving higher confidence with regards to the correctness of the code as well as ensuring strong parity of development, testing and production environments. This is done by using the [testcontainers-java](https://www.testcontainers.org/) library, a widely used (in industry and in the OSS community) Java library for instrumenting Docker containers from Java code for test purposes.

While launching a container instead of using an in-memory data has a certain overhead, one might argue the time and resource overhead is negligible when compared with the gains in confidence for integration correctness. Also, this works as a drop-in replacement by using the special [Testcontainers JDBC-Driver-Proxy](https://www.testcontainers.org/modules/databases/jdbc/), while at the same time mounting the data partition of the PostgreSQL container into host memory for better performance.

In addition, this allows to drop the need for maintaining a distinct set of flyway script for H2.

### Disclaimer

I am one of the core maintainers of the Testcontainers project. If you have any doubts or questions, I'd be happy to discuss them and to help in supporting the incorporation of Testcontainers for other integration testing use cases as well.


kiview:
I am wondering why the circle-ci build failed since I tested everything with a local Maven build first. However, I will try again after a `mvn clean` and investigate further.

bsideup:
@kiview because the build runs inside a container already and can't connect to Docker? :)
At least I don't see the CircleCI config changed, and I guess it needs to be:
https://www.testcontainers.org/supported_docker_environment/continuous_integration/circle_ci/

kiview:
Just the summarize the current status:
As @bsideup noted, the CircleCI config needed to be adjusted to use a VM instead of a Docker executor, so that Testcontainers is able to spin-up Docker containers for the integration tests.

However, there seems to be no simple way to specify the JDK version to be used in the VM (as opposed to the Docker executor approach) and it defaults to using JDK8, which is the reason why it is failing at the moment.

christian-kirschnick:
Closing due to inactivity - please re-open when you got a running version.

PR #541, opened 2020-06-11
Change SSL-related profiles to be secure by default

### LINKED ISSUE #482: SSL: Set a profile to willingly decrease security
Currently, we have the profiles `ssl-server`, `ssl-client-verification` and `ssl-client-postgres` which will enable `ssl` for certain situations. 

In context of secure-by-default, we should have it in the opposite way - you can set a profile to willingly decrease security.

pithumke:
# Summary
- Turn around all ssl-related profiles so that deployments can willingly deactivate security features instead of having to remember to activate them ("secure by default" mantra)
- Closes #482


MKusber:
I know the active profiles will be logged at startup. Is it possible to have dedicated `logger.warn` messages at startup to make it easier for monitoring to detect faulty configuration? I imagine something like:

```WARN: The service is started without TLS host verification enabled. This should never be used in PRODUCTION.```

File: services/distribution/src/main/java/app/coronawarn/server/services/distribution/Application.java [hunk 1]
```diff
@@ -58,4 +62,13 @@ public static void killApplication(ApplicationContext appContext) {
     logger.error("Application terminated abnormally.");
     System.exit(1);
   }
+
+  @Override
+  public void setEnvironment(Environment environment) {
+    List<String> profiles = Arrays.asList(environment.getActiveProfiles());
+    if (profiles.contains("disable-ssl-client-postgres")) {
+      logger.warn("The submission service is started with postgres connection TLS disabled. "
```
MKusber, line 70: You mean "distribution"?

services/submission/src/main/java/app/coronawarn/server/services/submission/ServerApplication.java
```diff
@@ -51,7 +63,24 @@ ProtobufHttpMessageConverter protobufHttpMessageConverter() {
...
   }
 
-  public static void main(String[] args) {
-    SpringApplication.run(ServerApplication.class);
+  @Override
+  public void setEnvironment(Environment environment) {
+    List<String> profiles = Arrays.asList(environment.getActiveProfiles());
+    if (profiles.contains("disable-ssl-server")) {
+      logger.warn("The submission service is started with endpoint TLS disabled. "
+          + "This should never be used in PRODUCTION!");
```
MKusber, line 71: Could we use a final String for the second sentence?

pithumke: Tried it, but if we go for simple String concatenation, then Sonar goes berserk. We could introduce Message formatters, but IMHO that's really not worth it. So I reverted.

pithumke:
Tested in local docker-compose and DEV landscape ✔️

PR #552, opened 2020-06-11
Enabling roles also for local development & docker setup

christian-kirschnick:
This PR moves the role definitions in the codebase, and enables local development to use the same setup as the deployed setup. Access permissions will now be handled via migration scripts.

Docker Compose is adjusted accordingly, so no further manual actions are needed.

For devs using their dedicated local postgres, you will need to run the SQL files for create-roles.sql & create-users.sql in the local-setup folder.

@MKusber


christian-kirschnick:
Please do not merge yet.

PR #559, opened 2020-06-12
DOC: Distribution Service/Spring Profiles

stevesap:


services/distribution/src/main/resources/application.yaml
```diff
@@ -9,20 +9,33 @@ logging:
 
 services:
   distribution:
+    # The name of the distribution output file.
     output-file-name: index
+    # The number of days to retain diagnosis keys for both database persistency layer and files stored on the object store.
     retention-days: 14
+    # The number of minutes which diagnosis keys are allowed to be expired for (since submission) before they can be distributed.
```
pithumke, line 16:
```suggestion
    # The number of minutes that diagnosis keys must have been expired for (since the end of the rolling interval window) before they can be distributed. Example: For a value of "expiry-policy-minutes: 120", the earliest point in time when a diagnosis key can be distributed is 26h after its "rolling-interval-start-number".
```

docs/DISTRIBUTION.md
```diff
@@ -1 +1,22 @@
...
+
+Spring profiles are used to apply distribution service configuration based on the running environment, determined by the active profile.
+
+You will find `.yaml` and `.xml` based profile-specific configuration files at [`/services/distribution/src/main/resources`](/services/distribution/src/main/resources).
+
+### Configuration Properties
+
+Please refer to the inline comments in the base `application.yaml` configuration file for further details.
+
+### Available Profiles
```
pithumke, line 13: Please update according to #541.

services/distribution/src/main/resources/application.yaml
stevesap: I updated the first sentence to the revised one but IMO the more extended explanation should be reserved for the relevant .md documentation section.

PR #563, opened 2020-06-12
Add Anchore container vulnerability scanning

jonico:
## Checklist

## Description

[cwa-server](https://github.com/jonico/cwa-server) uses two Docker images - the [submission](https://github.com/corona-warn-app/cwa-server/blob/master/services/submission/Dockerfile) and the [distribution image](https://github.com/corona-warn-app/cwa-server/blob/master/services/distribution/Dockerfile) - that may contain vulnerable components.
This PR introduces Docker container vulnerability scanning based on the [Open Source Anchore Scanning GitHub Action](https://anchore.com/blog/anchore-action-integrated-github-experience/) and report its findings into the Code scanning alert section of the security tab:

image

For each Docker image, one section is added. At the moment, 22 negligible findings (lowest warning level) show up - so the used base image (`gcr.io/distroless/java:11`) has no known security vulnerabilities 🎉 

The GitHub Action workflow triggers are designed that
* they run once a week to check whether there is any new vulnerability in the built Docker images
* they run whenever a Dockerfile is changed to check the differences
* they run whenever the GitHub Actions workflow file is changed itself

If this workflow should ever encounter a vulnerability with a severity > Medium, it will fail with an error; [the severity cut-off policy can be adjusted](https://github.com/anchore/scan-action#action-inputs) if needed.

### DIFF
--- a/.github/workflows/docker-images-analysis.yaml
+++ b/.github/workflows/docker-images-analysis.yaml
@@ -1,0 +1,55 @@
+name: Docker Images Analysis
+on:
+  push:
+    paths:
+      - '**/Dockerfile'
+      - '.github/workflows/docker-images-analysis.yaml'
+  schedule:
+      - cron: '0 5 * * 4'
+jobs:
+  distribution-docker-image-scan:
+      runs-on: ubuntu-latest
+      steps:
+      - uses: actions/checkout@v2
+      - name: Build the Docker image
+        run: docker build . --file services/distribution/Dockerfile --tag localbuild/testimage:latest
+      - uses: anchore/scan-action@master
+        with:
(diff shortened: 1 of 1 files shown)


stevesap:
Thanks for the proposal! Update: we already perform extensive vulnerability scanning of our Docker containers internally and will assess whether the proposed tool would yield any additional value.

michael-burwig:
Hi @jonico, thanks for putting your time and effort into this. As @stevesap pointed out, similar vulnerability scans are run as part of our internal pipeline, which we might publish more documentation on going forward.
At this point in time, the team would like to avoid the burden of triple-auditing. Hence, I will close this PR until we revisit the topic of externally visible scans again.

Thank you for your understanding.

PR #570, opened 2020-06-13
Fix fake delay updates

### LINKED ISSUE #474: Duration for fake-requests will be wrong if the server is under load
## Describe the bug

The [SubmissionController](https://github.com/cosee/cwa-server/blob/d0014f5c45f6021f84a0f64705f9a600b6476b9e/services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java#L120-L130) computes the duration of real submission requests in order to determine the test-results.

If the server is under load, the requests may get queued in the fork-join pool, which means that they will actually take longer than the measurement of the StopWatch, because the StopWatch doesn't start until the ForkJoinPool starts processing the job. The queue time is not respected.

This means that fake-requests will get a faster response.

This issue might become irrelevant depending on the solution of #399 

## Expected behaviour

Fake-requests should have a response-time similar to real requests.

## Steps to reproduce the issue

I have only looked at the code and not made any actual measurements. However, I am pretty sure it will behave as I described it.

## Possible Fix

The StopWatch should be started before the task is submitted to the ForkJoinPool.

pithumke:
### Before:
- Fake delay only gets updated on requests with valid TAN
- Scheduled executor becomes flooded with tasks and introduces additional delay
- This causes fake requests to be distinguishable from real requests under high load

Low load:
image

High load:
image

### After:
- Fake delay gets updated on all requests (valid or invalid TAN)
- Scheduled executor gets more threads to reduce additional delay
- Fake requests are indistinguishable from real requests even under high load

Low load:
image

High load:
image

### Note:
- Under extreme load (DDoS scenario), fake requests might still become slower than "real" requests, because also the 4 threads will eventually be flooded with tasks (although with this PR, that will happen a bit later). Not sure if there is a solution to this, other than request throttling or just outscaling the attack. Increasing the scheduled executor service thread pool size even more might improve this up to some point, but will also level off eventually.

Closes #474

### DIFF
--- a/services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java
+++ b/services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java
@@ -113,20 +113,21 @@ public class SubmissionController {
   private DeferredResult<ResponseEntity<Void>> buildRealDeferredResult(SubmissionPayload exposureKeys, String tan) {
     DeferredResult<ResponseEntity<Void>> deferredResult = new DeferredResult<>();
 
+    StopWatch stopWatch = new StopWatch();
+    stopWatch.start();
     try {
-      StopWatch stopWatch = new StopWatch();
-      stopWatch.start();
       if (!this.tanVerifier.verifyTan(tan)) {
         submissionControllerMonitor.incrementInvalidTanRequestCounter();
         deferredResult.setResult(buildTanInvalidResponseEntity());
       } else {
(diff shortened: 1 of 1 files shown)


nils4cosee:
@pithumke @christian-kirschnick I have some concerns with this change. Does this fall under the security-policy now, that the app is live?

christian-kirschnick:
Hey @nils4cosee, just drop me an e-mail describing your concerns. You can find my mail address in my profile.

PR #571, opened 2020-06-13
Doc: TLS

MKusber:
## Description

Add documentation to current TLS configuration and relevant Spring profiles.

### DIFF
--- a/docs/SECURITY.md
+++ b/docs/SECURITY.md
@@ -2,18 +2,20 @@
 
 ## TLS Configuration
 
-The cwa-server uses encrypted connections for external and internal communication. The main configuration for the external communication channel can be found [here](https://github.com/corona-warn-app/cwa-server/blob/36ac6f7cbcf9b230d6af8ddea96b387b47a3dee9/services/submission/src/main/resources/application.yaml#L76). The current settings will only allow TLS version 1.2. and 1.3. The offered ciphers are chosen in accordance to the recommendations in [BSI-TR-02102-2](https://www.bsi.bund.de/SharedDocs/Downloads/DE/BSI/Publikationen/TechnischeRichtlinien/TR02102/BSI-TR-02102-2.pdf).
-Internal communication is configured for instance [in the feign client](https://github.com/corona-warn-app/cwa-server/blob/master/services/submission/src/main/java/app/coronawarn/server/services/submission/verification/CloudFeignClientProvider.java).
-Database connections are established via TLS and are enabled for mutual authentication, see [here](https://github.com/corona-warn-app/cwa-server/blob/master/services/submission/src/main/resources/application.yaml#L38).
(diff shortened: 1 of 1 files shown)


docs/SECURITY.md
```diff
@@ -1 +1,19 @@
 # CWA-Server Security
+
+## TLS Configuration
+
+The cwa-server uses encrypted connections for external and internal communication. The main configuration for the external communication channel can be found [here](https://github.com/corona-warn-app/cwa-server/blob/36ac6f7cbcf9b230d6af8ddea96b387b47a3dee9/services/submission/src/main/resources/application.yaml#L76). The current settings will only allow TLS version 1.2. and 1.3. The offered ciphers are chosen in accordance to the recommendations in [BSI-TR-02102-2](https://www.bsi.bund.de/SharedDocs/Downloads/DE/BSI/Publikationen/TechnischeRichtlinien/TR02102/BSI-TR-02102-2.pdf).
```
stevesap, line 5: I would suggest to avoid permalinks if possible since then there's less chance of the documentation needing update. In this instance for example, I would have a relative link to the `application.yaml` and ask the reader to refer to the `server` configuration. 

Why? Let's say that in future we add some configuration properties and now `server` configuration is on line 80 instead. The permalink would still point them to the `server` configuration on line 76 of an older version of the file which the reader may mistakenly take to be the latest file and miss on out on some new properties that may have been added. With a relative link, they would always see the file in the current repo but will have to do a quick search for 'server' themselves.

stevesap, line 5: TLS version 1.2. and 1.3. -> TLS version 1.2 and 1.3.  (no dot after 1.2)

docs/SECURITY.md
```diff
@@ -1 +1,19 @@
...
+Internal communication is configured for instance [in the feign client](https://github.com/corona-warn-app/cwa-server/blob/master/services/submission/src/main/java/app/coronawarn/server/services/submission/verification/CloudFeignClientProvider.java).
```
stevesap, line 6: Also here I would switch to relative link

docs/SECURITY.md
```diff
@@ -1 +1,19 @@
...
+Database connections are established via TLS and are enabled for mutual authentication, see [here](https://github.com/corona-warn-app/cwa-server/blob/master/services/submission/src/main/resources/application.yaml#L38).
```
stevesap, line 7: Same here

docs/SECURITY.md
```diff
@@ -1 +1,19 @@
...
+
+## Security-Related Spring Profiles
+
+The TLS configuration previously described can be changed to enable certain test scenarios. It is important to realize that this degradation of security is only able by setting a Spring profile, while the default behavior will always use the most secure configuration. To prevent misconfiguration in production, multiple controls are implemented. This includes among others configuration-as-code deployments, explicit log messages for unsafe configurations and monitoring of deployed configurations.
```
stevesap, line 11: It is important to realize that this degradation of security is only able by setting a Spring profile -> It is important to realize that this degradation of security is only **possible** by setting a Spring profile

docs/SECURITY.md
```diff
@@ -1 +1,19 @@
...
+The relevant profiles are:
+
+* `disable-ssl-server` Disables SSL for the submission endpoint.
```
stevesap, line 14: Consider switching to a table format to be consistent with our other documentation of profiles (currently in README.md and being moved to overall documentation)

Issue #241, opened 2020-05-22
CodeStyle / Formatting.


indyaah:
* Source File: N/A
* Line(s): N/A
* Question:  Is there a shared code style adopted for the project? 

If not, I'd propose that https://github.com/google/google-java-format be adopted.

Pros: 
- Wider adoption
- [Solid tooling integration](https://github.com/google/google-java-format#third-party-integrations) (IntelliJ Plugin, Maven Plugin, Pre-commit hook support)
- Peace of mind while filing / reviewing PRs

Cons: 
- Possibly intrusive to people's personal setup.

Happy to take on and get maven plugin integrated if this sounds it would provide some value.

jonaskoepper:
To my understanding, this is already the case - Google Java Code Style was chosen.

- See https://github.com/corona-warn-app/cwa-server/tree/master/codestyle

- The according Checkstyle Maven plugin is also present: https://github.com/corona-warn-app/cwa-server/blob/master/pom.xml#L32

stevesap:
@michael-burwig can you confirm this is already the case?

CCFenner:
For simple formatting I can recommend [`.editorconfig`](https://editorconfig.org) which helps to take care about *dangling spaces*, *line breaks*, *indentation* and *indent type (space/tab)*.

IDE's like VSCode can then automatically stick to the defined rules with an additional plugin.

indyaah:
@stevesap : Yes, preferably automating code formatting before committing.  

Other option is to share IDE specific style configs and updating CONTRIBUTING.md to mention that your IDE needs to be setup to use them.

The issue with checkstyle is for some reason it's not really enforcing Google Java Style and it is reactive in nature. 

I have habit of autoformatting files after I have made changes to (IntelliJ with Google Code Style) and every single file I touch ends up having formatting changes that are 8-10x my actual change.

stevesap:
Hi @indyaah, we discussed and it sounds like it could be a nice addition to have an auto formatter integration to encourage wider adoption so feel free to proceed. 

Please keep in mind that as mentioned in the [CONTRIBUTING.md](https://github.com/corona-warn-app/cwa-server/blob/master/CONTRIBUTING.md#engaging-in-our-project), we are adhering to the Google Java Style, with the one exception of a max line width of 120 characters.

[...]

PR #573, opened 2020-06-13
Add validator for submissionTimestamp (closes #557)

### LINKED ISSUE #557: Add JPA range validator for submissionTimestamp of DiagnosisKey
DiagnosisKey uses a submissionTimestamp field of type long

https://github.com/corona-warn-app/cwa-server/blob/7059232b48d2507d142fe683fbebdda371a98652/common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/DiagnosisKey.java#L73 

the value is actually the *hours* since epoch but there are no sanity checks on the value.

Just from reading the code you could e.g. easily think it's seconds since epoch and use it the wrong way without noticing at runtime.

Proposal:

* Add a JPA ConstraintValidator similar to ValidRollingStartIntervalNumberValidator which checks the value range to be greater or equal zero and not in the future (current hour since epoch rounded up)
* Add field javadoc to submissionTimestamp: "hours since epoch"

Expected Benefits:

* discard invalid data
* less potential for programming errors

jsievers:
Similar to ValidRollingStartIntervalNumberValidator,
add ValidSubmissionTimestampValidator for field
submissionTimestamp.
Valid values are between 0 and the current hours since epoch.
Change field type in DiagnosisKeyBuilder to Long
object in order to distinguish a non-set default value
(null) from a negative value.

Fixes https://github.com/corona-warn-app/cwa-server/issues/557

### DIFF
--- a/common/persistence/src/test/java/app/coronawarn/server/common/persistence/domain/DiagnosisKeyBuilderTest.java
+++ b/common/persistence/src/test/java/app/coronawarn/server/common/persistence/domain/DiagnosisKeyBuilderTest.java
@@ -20,17 +20,19 @@
 
 package app.coronawarn.server.common.persistence.domain;
 
+import static app.coronawarn.server.common.persistence.domain.validation.ValidSubmissionTimestampValidator.SECONDS_PER_HOUR;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatCode;
 import static org.assertj.core.api.Assertions.catchThrowable;
-
 import app.coronawarn.server.common.persistence.exception.InvalidDiagnosisKeyException;
 import app.coronawarn.server.common.protocols.external.exposurenotification.TemporaryExposureKey;
 import com.google.protobuf.ByteString;
 import java.nio.charset.StandardCharsets;
+import java.time.Duration;
 import java.time.Instant;
 import java.time.LocalDate;
 import java.time.ZoneOffset;
+import java.time.temporal.ChronoUnit;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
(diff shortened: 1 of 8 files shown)


common/persistence/src/test/java/app/coronawarn/server/common/persistence/domain/DiagnosisKeyBuilderTest.java
```diff
@@ -218,12 +247,20 @@ private DiagnosisKey keyWithRiskLevel(int expTransmissionRiskLevel) {
         .withTransmissionRiskLevel(expTransmissionRiskLevel).build();
   }
 
+  private DiagnosisKey keyWithSubmissionTimestamp(long submissionTimestamp) {
```
michael-burwig, line 250: You could re-use ``DiagnosisKeyServiceTest.buildDiagnosisKeyForSubmissionTimestamp`` instead. Even though it might be placed a little awkwardly there, I assume we can extract it to somewhere else in the future.

common/persistence/src/test/java/app/coronawarn/server/common/persistence/domain/DiagnosisKeyBuilderTest.java
```diff
@@ -20,17 +20,19 @@
 
 package app.coronawarn.server.common.persistence.domain;
 
+import static app.coronawarn.server.common.persistence.domain.validation.ValidSubmissionTimestampValidator.SECONDS_PER_HOUR;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatCode;
 import static org.assertj.core.api.Assertions.catchThrowable;
-
```
michael-burwig, line 26: At least in my intelliJ (due to the entries in ``.editorconfig``) this line would get reverted once the formatter is applied. My apologies for the somewhat intransparent situation here.

common/persistence/src/test/java/app/coronawarn/server/common/persistence/domain/DiagnosisKeyBuilderTest.java
jsievers: thanks, I hadn't noticed that. I extracted and reused the methods from the package private test class to the already existing public helper class now.

common/persistence/src/test/java/app/coronawarn/server/common/persistence/domain/DiagnosisKeyBuilderTest.java
jsievers: yes I saw https://github.com/corona-warn-app/cwa-server/issues/241 but didn't get the .editorconfig to work with the eclipse formatter . I fixed it manually now

jsievers:
thanks for reviewing!

PR #590, opened 2020-06-15
Add manual Log4J shutdown hook (close #589)

### LINKED ISSUE #589: Log4j improper shutdown
## Current Behaviour

On Distribution service shutdown, the following log entry appears, indicating Log4j did not shutdown properly. Also, it is logged using the fallback `SimpleLogger` which uses a different timestamp date format and is not consistent with other log entries and therefore may not be picked up by monitoring tools.
 
`2020-06-12 09:05:23,582 SpringContextShutdownHook WARN Unable to register Log4j shutdown hook because JVM is shutting down. Using SimpleLogger`

This is due to an [unresolved bug](https://issues.apache.org/jira/browse/LOG4J2-658) with Log4j.

## Possible Fix

As described in https://stackoverflow.com/questions/30120330/log4j2-unable-to-register-shutdown-hook-because-jvm-is-shutting-down/33134173, a workaround is:

- Disable Log4j's automatic shutdown hook
- Register a manual shutdown hook instead

stevesap:
## Checklist

## Description

Add manual Log4J shutdown hook to resolve improper Log4j shutdown (see #589)

### DIFF
--- a/services/distribution/src/main/java/app/coronawarn/server/services/distribution/Application.java
+++ b/services/distribution/src/main/java/app/coronawarn/server/services/distribution/Application.java
@@ -59,7 +59,7 @@ public class Application implements EnvironmentAware {
   public static void main(String[] args) {
     SpringApplication.run(Application.class);
 
-    // Manual shutdown hook needed to avoid Log4j shutdown issues (see cwa-server/#)
+    // Manual shutdown hook needed to avoid Log4j shutdown issues (see cwa-server/#589)
     Runtime.getRuntime().addShutdownHook(new Thread(() -> {
       try {
         logger.info("Shutting down log4j2.");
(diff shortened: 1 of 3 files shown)


services/distribution/src/main/java/app/coronawarn/server/services/distribution/Application.java
```diff
@@ -50,8 +51,23 @@
...
+  /**
+   * The application's entry point.
+   *
+   * @param args command-line arguments for the application (unused).
+   */
   public static void main(String[] args) {
     SpringApplication.run(Application.class);
+
+    // Manual shutdown hook needed to avoid Log4j shutdown issues (see cwa-server/#589)
+    Runtime.getRuntime().addShutdownHook(new Thread(() -> {
```
christian-kirschnick, line 63: Can you try `@PreDestroy` annotation instead? Would make it more "springy"

PR #591, opened 2020-06-15
Multi-line Exception logging: Replace all CR, LF and CRLF

stevesap:
## Checklist

## Description

We replace all EOL's with `\u2028` in logged Exceptions. This is in order to avoid multiple log entries in monitoring tools for error message with multi-lined exception.

This PR makes sure to replace all possible combinations of EOL's - CR, LF and CRLF


stevesap:
@MKusber

PR #598, opened 2020-06-16
Fix/appconfig validation

michael-burwig:
- Fixes bug where app configurations could be written even if validation failed at runtime. (rather theoretic, since unit tests would fail in this case)
- Adds unit test for ``AppConfigurationDirectory``
- Corrects JavaDoc in ``AppConfigurationDirectory``
- Extracts test _helper_ method
- Refactors ``AppConfigurationProvider`` into Configuration that provides ``ApplicationConfiguration`` bean

### DIFF
--- a/services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/appconfig/ApplicationConfigurationProvider.java
+++ b/services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/appconfig/ApplicationConfigurationProvider.java
@@ -30,28 +30,11 @@ import app.coronawarn.server.common.protocols.internal.ApplicationConfiguration;
  *   <li>Risk Score Classification</li>
  *   <li>App Config, e.g. minimum risk threshold</li>
  * </ul>
- *
- * <p>The application config is fetched from the master-config folder.</p>
  */
 public class ApplicationConfigurationProvider {
 
-  /**
-   * The location of the exposure configuration master file.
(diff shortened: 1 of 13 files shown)


services/distribution/src/main/resources/application.yaml
```diff
@@ -79,6 +79,8 @@ services:
       max-number-of-failed-operations: 5
       # The ThreadPoolTaskExecutor's maximum thread pool size.
       max-number-of-s3-threads: 4
+    # The location of the app configuration parameters file on the classpath
```
stevesap, line 82:
```suggestion
    # The location of the app configuration parameters file on the classpath.
```
For consistency

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/appconfig/ApplicationConfigurationProvider.java
```diff
@@ -30,30 +30,13 @@
...
  * </ul>
- *
- * <p>The application config is fetched from the master-config folder.</p>
  */
 public class ApplicationConfigurationProvider {
 
-  /**
-   * The location of the exposure configuration master file.
-   */
-  public static final String MASTER_FILE = "master-config/app-config.yaml";
```
christian-kirschnick, line 41: Not sure about this change tbh - I don't see any reason why somebody would want to have that file path configurable at all.

Issue #237, opened 2020-05-21
circleci sonarQube scan currently fails for forked PR builds


johanneseschrig:
## Describe the bug

PRs from forks will fail with `[ERROR] Failed to execute goal org.sonarsource.scanner.maven:sonar-maven-plugin:3.7.0.1746:sonar (default-cli) on project server: Parameter 'sonar.pullrequest.branch' is mandatory for a pull request analysis`. 

Needs to be fixed.

johanneseschrig:
Consider: https://github.com/corona-warn-app/cwa-server/pull/236#issuecomment-632226214

jonaskoepper:
Seems this is resolved? My PR, based on a fork, meanwhile builds: https://github.com/corona-warn-app/cwa-server/pull/239

johanneseschrig:
Hi Jonas! Unfortunately we currently just work around it with https://github.com/corona-warn-app/cwa-server/pull/236. This way we don't block any PRs, but sonar still doesn't run on community PRs atm.

CCFenner:
Recent comment from Sonar officials in https://community.sonarsource.com/t/how-to-use-sonarcloud-with-a-forked-repository-on-github/7363/8


> > Are there any news? The Issue https://jira.sonarsource.com/browse/MMF-1371 is not so active :slight_smile:

> Not yet! I recommend to check again after 4-6 months.

//fyi @jhagestedt

CCFenner:
A workaround for this could be a chat-ops-like workflow. The following would be needed:

- a workflow that listens to *issue comments*
  - if specific comment (e.g. `/sonar`) is posted on a fork pr push the changes from the fork to a dedicated branch on the upstream repository
- a workflow that runs sonar *on push* to the dedicated branch

This would require a PAT of a personal or technical user in repository secrets with repo write permissions as the default `secrets.GITHUB_TOKEN` is not sufficient.

Would this be a way to go?

PR #599, opened 2020-06-16
Add missing blanks in warn messages

stweil:
This fixes three alerts reported by LGTM:
https://lgtm.com/projects/g/corona-warn-app/cwa-server/?mode=list

Signed-off-by: Stefan Weil <sw@weilnetz.de>

### DIFF
--- a/services/submission/src/main/java/app/coronawarn/server/services/submission/ServerApplication.java
+++ b/services/submission/src/main/java/app/coronawarn/server/services/submission/ServerApplication.java
@@ -73,18 +73,18 @@ public class ServerApplication implements EnvironmentAware {
     }
     if (profiles.contains("disable-ssl-client-postgres")) {
       logger.warn(
-          "The submission service is started with postgres connection TLS disabled. This should never be used in"
-              + "PRODUCTION!");
+          "The submission service is started with postgres connection TLS disabled. "
+              + "This should never be used in PRODUCTION!");
     }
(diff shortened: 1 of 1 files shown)


michael-burwig:
Hi @stweil, looks like you (and LGTM) is right here wrt the whitespaces.

On a different note, I remember us putting ``+ "This should never be used in PRODUCTION!"`` deliberately *not* onto new lines for each warn message, but breaking the string at max line width instead. This was because Sonarcloud was complaining about code duplication otherwise and this seemed like the most straightforward fix.

stweil:
Then Sonarcloud should be fixed. Don't write badly formatted code to satisfy buggy tools.

But maybe there is no problem, because Sonarcloud is clever enough to see that it is no code duplication. I'd suggest to simply try it.

stevesap:
I checked locally with SonarLint (due to #237) and it is giving [String literals should not be duplicated](https://rules.sonarsource.com/go/RSPEC-1192#:~:text=Duplicated%20string%20literals%20make%20the,updated%20in%20a%20single%20place.) issue.

Use a string constant instead?

pithumke:
See https://github.com/corona-warn-app/cwa-server/pull/541#discussion_r439301614

stweil:
> I checked locally with SonarLint (due to #237) and it is giving String literals should not be duplicated issue.

As Java is not fully supported by Sonar, I'd disable the test for duplicate string literals.

PR #605, opened 2020-06-19
Clearing up possible misunderstanding of full architecture vs cwa server

christian-kirschnick:
This PR adds more information of the overall solution architecture in the architecture document.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,26 +170,11 @@ Distribution Service      | [services/distribution/api_v1.json](https://github.c
 
 ### Distribution
 
-Profile                       | Effect
-------------------------------|-------------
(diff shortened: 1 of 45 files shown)


JudgeSi:
I think it would be nice if the [readme](https://github.com/corona-warn-app/cwa-server/blob/master/README.md) would point to this document (please correct me if I am wrong). At the moment this section is a little hidden.
I think a great place would be the architecture section of the readme.

christian-kirschnick:
@JudgeSi It is already linked in the Architecture overview section of the readme:

image

Guess it is still too hidden :) Maybe you can create a PR to change the readme in a way which will make it more visible.

Issue #108, opened 2020-05-15
Aggregate File Creation: Skip packages if payload too small


christian-kirschnick:
Change the aggregate file creation logic in a way, in which packages are skipped in case the payload (amount of keys) is too small. The threshold should be configurable, and defaulted to 50 for now.

MikeJayDee:
**Alternative suggestion: Pad packages with random dummy keys if too small**

Thinking a bit further into the hoped for future, when new infections are becoming rare, this might lead to unnecessary (possibly quite long) delays in notifications being sent. Could one instead use random dummy keys to pad every package that is too small to maybe 50 or 100 keys (or possibly padding to a randomised amount so an outsider can't even determine whether there any dummies are in the set).

This would impact required data volumes marginally - in the worst case, if 75 dummy keys are added every hour, this would represent only 50 kiB per day (or 3% of your bandwidth example in the architecture document) if I did not miscalculate. Since this would only occur in a low load situation anyway, this should be acceptable.

christian-kirschnick:
Threhold is now: 140

christian-kirschnick:
I like the padding idea - @MKusber do you think we can do this from a DPP/security perspective?

pithumke:
Hey @MikeJayDee, after some investigation, @michael-burwig and me decided that we will go for the "shifting" approach instead of the "padding" approach for this release. For clarification:

**"Shifting" approach:**
- If a package is too small (number of keys submitted in a timeframe is below a configurable threshold), then it will not be distributed (there will be zero keys in that package)
- The keys that would have been distributed in that run are "shifted" to the next distribution run
- Rinse and repeat until a package can be assembled that contains enough keys (above the configurable threshold)

**"Padding" approach:**
- If a package is too small, then it will be padded with random / fake keys until the threshold is matched (or exceeded)

As you have rightly pointed out, the main disadvantage of the "shifting" approach is that it introduces an artificial delay in the distribution of keys. This means that people will not be warned about a potential COVID exposure as quickly as possible, but that the CWA server will wait with the distribution until enough submissions have been made. Especially as the number of submissions decreases, this delay will become significant.

[...]

PR #609, opened 2020-06-20
Add random key padding

pithumke:
### Summary:
- Generate a fixed number of random diagnosis keys for every real submitted key
  - Random diagnosis keys share all properties of the real key, except for the `key_data`
- Add configurable `random-key-padding-multiplier` property
- Improves https://github.com/corona-warn-app/cwa-server/issues/108
- Related to https://github.com/corona-warn-app/cwa-documentation/issues/236
- Addresses CCC "Prüfstein" Point 9 "Unlinkability" without introducing additional delay:
  - https://github.com/corona-warn-app/cwa-documentation/blob/master/pruefsteine.md#unlinkability

Documentation for the `random-key-padding-multiplier` property from [application.yaml](https://github.com/corona-warn-app/cwa-server/pull/609/files#diff-561c3bb602ad52bed235b8f76cd8a437R23):
```yaml
# The number of keys to save to the DB for every real submitted key.
# Example: If the 'random-key-padding-multiplier' is set to 10, and 5 keys are being submitted,
# then the 5 real submitted keys will be saved to the DB, plus an additional 45 keys with
# random 'key_data'. All properties, besides the 'key_data', of the additional keys will be
# identical to the real key.
```

### To be decided
- [ ] What value should we set as the `random-key-padding-multiplier`? (currently set to `10`)
- [ ] Can we remove (or at least reduce) the "shifting policy threshold"? (currently set to `140`)

### DIFF
--- a/services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java
+++ b/services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java
@@ -29,8 +29,11 @@ import app.coronawarn.server.services.submission.monitoring.SubmissionMonitor;
 import app.coronawarn.server.services.submission.validation.ValidSubmissionPayload;
 import app.coronawarn.server.services.submission.verification.TanVerifier;
 import io.micrometer.core.annotation.Timed;
+import java.security.SecureRandom;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpStatus;
@@ -59,6 +62,7 @@ public class SubmissionController {
   private final DiagnosisKeyService diagnosisKeyService;
   private final TanVerifier tanVerifier;
   private final Integer retentionDays;
+  private final Integer randomKeyPaddingMultiplier;
   private final FakeDelayManager fakeDelayManager;
 
   SubmissionController(
@@ -69,6 +73,7 @@ public class SubmissionController {
     this.submissionMonitor = submissionMonitor;
     this.fakeDelayManager = fakeDelayManager;
     retentionDays = submissionServiceConfig.getRetentionDays();
+    randomKeyPaddingMultiplier = submissionServiceConfig.getRandomKeyPaddingMultiplier();
   }
 
   /**
@@ -130,6 +135,26 @@ public class SubmissionController {
       }
     }
 
-    diagnosisKeyService.saveDiagnosisKeys(diagnosisKeys);
+    diagnosisKeyService.saveDiagnosisKeys(padDiagnosisKeys(diagnosisKeys));
+  }
+
+  private List<DiagnosisKey> padDiagnosisKeys(List<DiagnosisKey> diagnosisKeys) {
+    List<DiagnosisKey> paddedDiagnosisKeys = new ArrayList<>();
+    diagnosisKeys.forEach(diagnosisKey -> {
+      paddedDiagnosisKeys.add(diagnosisKey);
+      IntStream.range(1, randomKeyPaddingMultiplier)
+          .mapToObj(index -> {
+            byte[] randomKeyData = new byte[16];
+            new SecureRandom().nextBytes(randomKeyData);
+            return DiagnosisKey.builder()
+                .withKeyData(randomKeyData)
+                .withRollingStartIntervalNumber(diagnosisKey.getRollingStartIntervalNumber())
+                .withTransmissionRiskLevel(diagnosisKey.getTransmissionRiskLevel())
+                .withRollingPeriod(diagnosisKey.getRollingPeriod())
+                .build();
+          })
+          .forEach(paddedDiagnosisKeys::add);
+    });
+    return paddedDiagnosisKeys;
   }
 }
--- a/services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java
+++ b/services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java
@@ -32,7 +32,6 @@ import io.micrometer.core.annotation.Timed;
 import java.security.SecureRandom;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
--- a/services/submission/src/main/resources/application.yaml
+++ b/services/submission/src/main/resources/application.yaml
@@ -15,6 +15,12 @@ services:
     fake-delay-moving-average-samples: 10
     # The retention threshold for acceptable diagnosis keys during submission.
     retention-days: 14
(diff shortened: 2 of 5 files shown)


services/submission/src/test/java/app/coronawarn/server/services/submission/controller/SubmissionControllerTest.java
```diff
@@ -253,17 +253,33 @@ private TemporaryExposureKey createOutdatedKey() {
   }
 
   private void assertElementsCorrespondToEachOther
-      (Collection<TemporaryExposureKey> submittedKeys, Collection<DiagnosisKey> keyEntities) {
-    Set<DiagnosisKey> expKeys = submittedKeys.stream()
-        .map(aSubmittedKey -> DiagnosisKey.builder().fromProtoBuf(aSubmittedKey).build())
+      (Collection<TemporaryExposureKey> actTemporaryExposureKeys, Collection<DiagnosisKey> expDiagnosisKeys) {
```
pithumke, line 256: I think I mixed up actual/expected and saved/submitted in the variable naming and testing logic. Will fix this tomorrow.

services/submission/src/test/java/app/coronawarn/server/services/submission/controller/SubmissionControllerTest.java
```diff
@@ -253,17 +253,33 @@ private TemporaryExposureKey createOutdatedKey() {
...
+      List<DiagnosisKey> expKeyEntities = expDiagnosisKeys.stream()
+          .filter(expDiagnosisKey ->
+              expDiagnosisKey.getRollingPeriod() == actDiagnosisKey.getRollingPeriod())
+          .filter(expDiagnosisKey ->
+              expDiagnosisKey.getTransmissionRiskLevel() == actDiagnosisKey.getTransmissionRiskLevel())
+          .filter(expDiagnosisKey ->
+              expDiagnosisKey.getRollingStartIntervalNumber() == actDiagnosisKey.getRollingStartIntervalNumber())
+          .collect(Collectors.toList());
+
+      assertThat(expKeyEntities).hasSize(10);
```
pithumke, line 273: Doesn't make any sense.

services/submission/src/test/java/app/coronawarn/server/services/submission/controller/SubmissionControllerTest.java
```diff
@@ -253,17 +253,33 @@ private TemporaryExposureKey createOutdatedKey() {
...
+      assertThat(expKeyEntities.stream().filter(expKeyEntity ->
```
pithumke, line 274: Check if something like a `oneMatches` assertion is available.

pithumke: Does not exist in AssertJ

services/submission/src/test/java/app/coronawarn/server/services/submission/controller/SubmissionControllerTest.java
pithumke: Fixed

services/submission/src/test/java/app/coronawarn/server/services/submission/controller/SubmissionControllerTest.java
```diff
@@ -252,18 +252,36 @@ private TemporaryExposureKey createOutdatedKey() {
...
         .collect(Collectors.toSet());
 
-    assertThat(keyEntities)
-        .withFailMessage("Number of submitted keys and generated key entities don't match.")
-        .hasSameSizeAs(expKeys);
-    keyEntities.forEach(anActKey -> assertThat(expKeys)
-        .withFailMessage("Key entity does not correspond to a submitted key.")
-        .contains(anActKey)
-    );
+    assertThat(savedDiagnosisKeys).hasSize(submittedDiagnosisKeys.size() * 10);
```
michael-burwig, line 262: magic `10` should rather be read from the config

services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java
```diff
@@ -130,6 +134,26 @@ public void persistDiagnosisKeysPayload(SubmissionPayload protoBufDiagnosisKeys)
...
+    diagnosisKeyService.saveDiagnosisKeys(padDiagnosisKeys(diagnosisKeys));
+  }
+
+  private List<DiagnosisKey> padDiagnosisKeys(List<DiagnosisKey> diagnosisKeys) {
+    List<DiagnosisKey> paddedDiagnosisKeys = new ArrayList<>();
+    diagnosisKeys.forEach(diagnosisKey -> {
+      paddedDiagnosisKeys.add(diagnosisKey);
+      IntStream.range(1, randomKeyPaddingMultiplier)
+          .mapToObj(index -> {
+            byte[] randomKeyData = new byte[16];
```
michael-burwig, line 146: Consider extracting _random key_ creation logic

pinkit-cwa:
@pithumke Is there a specific reason why you chose to just add random TEKs for every TEK instead of rounding up the number of TEKs to the next 10 or maybe 32?
I am not into statistics, but [these guys](https://down.dsg.cs.tcd.ie/tact/tek-counts/) seem to be, and comment:
> For some reason the German server publishes 10 keys for every one really uploaded. (See this [github issue](https://github.com/corona-warn-app/cwa-server/pull/609).) I don't really buy that as a privacy win TBH - just rounding up to a multiple of 10 would be fine, but at least I think I now understand the numbers.

pithumke:
Hello @pinkit-cwa,
as outlined in [this comment](https://github.com/corona-warn-app/cwa-server/issues/108#issuecomment-636541556) and [this comment](https://github.com/corona-warn-app/cwa-server/issues/108#issuecomment-647716745), we chose this approach, because we want to avoid DPP problems arising from statistical analysis of random TEKs while still keeping the 140 threshold required by the RKI (but reaching it faster). Simply duplicating all real keys N times and only forging the `key_data` portion, avoids this potential problem, was quick and easy to implement and was signed off by the RKI. If we were to instead "round up" the total number of TEKs, we would need to randomly generate full TEKs from scratch (including the transmission risk level and the rolling start interval number), potentially risking those fake keys to be identifiable by statistical analysis.

services/submission/src/main/resources/application.yaml
```diff
@@ -15,6 +15,12 @@ services:
     fake-delay-moving-average-samples: 10
     # The retention threshold for acceptable diagnosis keys during submission.
     retention-days: 14
+    # The number of keys to save to the DB for every real submitted key.
+    # Example: If the 'random-key-padding-multiplier' is set to 10, and 5 keys are being submitted,
+    # then the 5 real submitted keys will be saved to the DB, plus an additional 45 keys with
+    # random 'key_data'. All properties, besides the 'key_data', of the additional keys will be
+    # identical to the real key.
+    random-key-padding-multiplier: ${RANDOM_KEY_PADDING_MULTIPLIER:1}
```
hilmarf, line 23: @ChristopherSchmitz - luckily Pit made this padding already configurable and the default is only 1. So as long as security/DPP has no other concerns, we can simply change the value, if required.

PR #619, opened 2020-06-23
Enchantments - Increase test coverage

robertoduessmann:
## Checklist

## Description

Increase project test coverage:
1) Added a new scenario to `VerificationServiceHealthIndicatorTest.java` to cover the case where no exception is thrown and the app is also healthy.
2) I realise there was coverage missing in the `ValidSubmissionPayload.java` for the check `Objects.isNull(exposureKeys)`.
But in fact it can't be null since it's a proto list, and by default it will be converted in a empty list. I suggest to remove the validation in this case to keep the code cleaner.
From [proto3 documentation](https://developers.google.com/protocol-buffers/docs/proto3#default):
> The default value for repeated fields is empty (generally an empty list in the appropriate language).


michael-burwig:
@robertoduessmann, thank you for your contribution!

PR #621, opened 2020-06-24
Add unit tests for export.bin.checksum

### LINKED ISSUE #606: CWA Hash should also incorporate the signing information
We are currently using the `export.bin` payload in order to compute the checksum. This approach does not reflect the signing information, which is part of the `export.sig`. Therefore, if the key would be rolled at some point in time, the cwa hash will not change, and hence the system will not re-distribute the files with the new signature information.

Therefore, the key cwa hash should also reflect the key, which was used for signing they `export.bin` payload.

Current is:
https://github.com/corona-warn-app/cwa-server/blob/cd4eb1af6f8f60ae2b97759cbdeaaead84a20f13/services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/structure/file/FileOnDiskWithChecksum.java#L75-L80

mibrasap:
## Checklist

## Description

Add unit tests for export.bin.checksum. Checking if checksum is calculated deterministically and incorporates changes on the signature's verification key version.

closes #606

### DIFF
--- a/services/distribution/src/test/java/app/coronawarn/server/services/distribution/assembly/diagnosiskeys/structure/file/TemporaryExposureKeyExportFileTest.java
+++ b/services/distribution/src/test/java/app/coronawarn/server/services/distribution/assembly/diagnosiskeys/structure/file/TemporaryExposureKeyExportFileTest.java
@@ -1,3 +1,23 @@
+/*-
+ * ---license-start
+ * Corona-Warn-App
+ * ---
+ * Copyright (C) 2020 SAP SE and all other contributors
+ * ---
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ---license-end
+ */
+
 package app.coronawarn.server.services.distribution.assembly.diagnosiskeys.structure.file;
 
 import app.coronawarn.server.common.persistence.domain.DiagnosisKey;
--- a/services/distribution/src/test/java/app/coronawarn/server/services/distribution/assembly/diagnosiskeys/structure/file/TemporaryExposureKeyExportFileTest.java
+++ b/services/distribution/src/test/java/app/coronawarn/server/services/distribution/assembly/diagnosiskeys/structure/file/TemporaryExposureKeyExportFileTest.java
@@ -1,0 +1,108 @@
+package app.coronawarn.server.services.distribution.assembly.diagnosiskeys.structure.file;
+
+import app.coronawarn.server.common.persistence.domain.DiagnosisKey;
+import app.coronawarn.server.services.distribution.assembly.structure.WritableOnDisk;
+import app.coronawarn.server.services.distribution.assembly.structure.directory.Directory;
+import app.coronawarn.server.services.distribution.assembly.structure.directory.DirectoryOnDisk;
+import app.coronawarn.server.services.distribution.assembly.structure.util.ImmutableStack;
+import app.coronawarn.server.services.distribution.config.DistributionServiceConfig;
+import app.coronawarn.server.services.distribution.config.DistributionServiceConfig.Signature;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.Arrays;
+import java.util.Set;
+import org.junit.Rule;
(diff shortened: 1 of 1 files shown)


services/distribution/src/test/java/app/coronawarn/server/services/distribution/assembly/diagnosiskeys/structure/file/TemporaryExposureKeyExportFileTest.java
```diff
@@ -0,0 +1,128 @@
...
+  void testChecksumChangesWhenPrivateKeyIsModified() throws IOException {
+    File outputFile1 = createFile();
+    byte[] checksum1 = readChecksumFile(outputFile1);
+
+    modifySignaturePrivateKeyVersion();
+
+    File outputFile2 = createFile();
+    byte[] checksum2 = readChecksumFile(outputFile2);
+
+    assertThat(!(Arrays.equals(checksum1, checksum2)));
```
michael-burwig, line 91: Better use of assertions:
```suggestion
    assertThat(checksum1).isNotEqualTo(checksum2);
```

services/distribution/src/test/java/app/coronawarn/server/services/distribution/assembly/diagnosiskeys/structure/file/TemporaryExposureKeyExportFileTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+  @Test
+  void testChecksumIsDeterministic() throws IOException {
+    File outputFile1 = createFile();
+    byte[] checksum1 = readChecksumFile(outputFile1);
+
+    File outputFile2 = createFile();
+    byte[] checksum2 = readChecksumFile(outputFile2);
+
+    assertThat(Arrays.equals(checksum1, checksum2));
```
michael-burwig:
```suggestion
    assertThat(checksum1).isEqualTo(checksum2);
```

services/distribution/src/test/java/app/coronawarn/server/services/distribution/assembly/diagnosiskeys/structure/file/TemporaryExposureKeyExportFileTest.java
```diff
@@ -0,0 +1,128 @@
...
+import org.junit.rules.TemporaryFolder;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.boot.test.context.ConfigFileApplicationContextInitializer;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.annotation.DirtiesContext.ClassMode;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+import static org.assertj.core.api.Assertions.assertThat;
```
michael-burwig, line 48: Consider moving before non-static imports for consistency

services/distribution/src/test/java/app/coronawarn/server/services/distribution/assembly/diagnosiskeys/structure/file/TemporaryExposureKeyExportFileTest.java
```diff
@@ -0,0 +1,128 @@
...
+    File outputFile2 = createFile();
+    byte[] checksum2 = readChecksumFile(outputFile2);
+
+    assertThat(!(Arrays.equals(checksum1, checksum2)));
+  }
+
+  private TemporaryExposureKeyExportFile createTemporaryExposureKeyExportFile() {
+    return TemporaryExposureKeyExportFile.fromDiagnosisKeys(
+        Set.of(
+            DiagnosisKey.builder()
```
michael-burwig, line 97: Consider re-use of ``Helpers.buildDiagnosisKeyForSubmissionTimestamp`` or similar

services/distribution/src/test/java/app/coronawarn/server/services/distribution/assembly/diagnosiskeys/structure/file/TemporaryExposureKeyExportFileTest.java
mibrasap: Checksum1 and checksum2 are arrays; `isEqualTo` and `isNotEqualTo` is checking for referential equality, which we do not want. We need a check for equal content of the arrays' content.

michael-burwig: I assume that ``isEqualTo`` is employing ``org.assertj.core.internal.ObjectArrayElementComparisonStrategy`` here, which performs a _propper_ array comparison.
That is also what I could observed when trying it.

michael-burwig: Okay, actually it uses ``Arrays.deepEquals`` internally.

PR #625, opened 2020-06-25
OSP is OpenStack Platform, OpenShift is OCP

dabelenda:
## Checklist

## Description

Since OpenShift is OCP/OKD, correct the acronym in the architecture document.


michael-burwig:
Thanks, looks about right. I will check if this abbreviation used in other places as well.

PR #626, opened 2020-06-25
Index Files: Incorrect mime type closes #608

### LINKED ISSUE #608: Index Files: Incorrect mime type
Index files like https://svc90.main.px.t-online.de/version/v1/diagnosis-keys/country/DE/date currently respond with mime type `application/octet-stream`. Instead, the correct mime type for JSON should be used.

hilmarf:
## Description

#608 

Based upon the s3Key, it figures out what to use as content-type header.


hilmarf:
new PR 627 created with better branch name

PR #631, opened 2020-06-29
Replace data-jpa with data-jdbc

michael-burwig:
Given our current implementation and requirements, the more lightweight Spring Data JDBC dependency can provide feature parity without increasing code complexity.

- Switches from Spring Data JPA to Data JDBC.
- Enables flyway migration for unit tests
- Adds Spring-AOP as a dependency to distribution (previously provided transitively).

See Spring Data JDBC documentation for an introduction:
- [Spring Data JDBC - Reference Documentation](https://docs.spring.io/spring-data/jdbc/docs/2.0.1.RELEASE/reference/html/#reference)
- [Spring Data JDBC - References and Aggregates](https://spring.io/blog/2018/09/24/spring-data-jdbc-references-and-aggregates)

### DIFF
--- a/services/distribution/src/main/resources/application.yaml
+++ b/services/distribution/src/main/resources/application.yaml
@@ -84,9 +84,6 @@ spring:
   main:
     web-application-type: NONE
   # Postgres configuration
-  jpa:
-    hibernate:
-      ddl-auto: validate
   flyway:
     enabled: true
     locations: classpath:/db/migration, classpath:/db/specific/{vendor}
--- a/services/distribution/src/test/resources/application.yaml
+++ b/services/distribution/src/test/resources/application.yaml
@@ -54,8 +54,10 @@ services:
 spring:
   main:
     banner-mode: off
-  flyway:
-    enabled: true
-  jpa:
-    hibernate:
-      ddl-auto: validate
+  datasource:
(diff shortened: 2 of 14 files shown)


services/distribution/src/test/resources/application.yaml
```diff
@@ -54,8 +54,10 @@ services:
...
-  flyway:
-    enabled: true
-  jpa:
-    hibernate:
-      ddl-auto: validate
+  datasource:
+    url: jdbc:h2:mem:test;MODE=PostgreSQL
+    driverClassName: org.h2.Driver
+    platform: h2
+  test:
```
michael-burwig, line 61: Add comment

services/distribution/src/main/resources/application.yaml
```diff
@@ -84,9 +84,6 @@ spring:
   main:
     web-application-type: NONE
   # Postgres configuration
-  jpa:
-    hibernate:
-      ddl-auto: validate
```
pithumke, line 89: Does some other component perform a schema validation now?

michael-burwig: To some extent you could say that Flyway is ensuring the presence of the expected schema at startup. Beyond that we will have to, for example, rely on unit tests to detect bugs.

PR #642, opened 2020-07-01
Fix/attenuation value range

### LINKED ISSUE #638: [MINOR] Incorrect value range in attenuation risk parameter
## Describe the bug
The `AttenuationRiskParameters` proto message (see [here](https://github.com/corona-warn-app/cwa-server/blob/master/common/protocols/src/main/proto/app/coronawarn/server/common/protocols/internal/risk_score_parameters.proto)) is defined as follows:
```proto3
message AttenuationRiskParameters {
  RiskLevel gt_73_dbm = 1;            // A > 73 dBm, lowest risk
  RiskLevel gt_63_le_73_dbm = 2;      // 63 < A <= 73 dBm
  RiskLevel gt_51_le_63_dbm = 3;      // 51 < A <= 63 dBm
  RiskLevel gt_33_le_51_dbm = 4;      // 33 < A <= 51 dBm
  RiskLevel gt_27_le_33_dbm = 5;      // 27 < A <= 33 dBm
  RiskLevel gt_15_le_27_dbm = 6;      // 15 < A <= 27 dBm
  RiskLevel gt_10_le_15_dbm = 7;      // 10 < A <= 15 dBm
  RiskLevel lt_10_dbm = 8;            // A <= 10 dBm, highest risk
}
```

There is no value range that includes an attenuation of **exactly** 10dBm.

## Expected behaviour
According to the [Google/Apple spec](https://developer.apple.com/documentation/exposurenotification/enexposureconfiguration), the last value range should be "less than or equal to 10dBm" (`le_10_dbm`):

image

## Possible Fix
- Correct the last attenuation value range according to spec
- Update dependent code/configuration respectively ([here](https://github.com/corona-warn-app/cwa-server/blob/b8afdce9a3bc8fd927fe4ec9be2e910ee8f1635d/services/distribution/src/test/resources/objectstore/publisher/examplefile), [here](https://github.com/corona-warn-app/cwa-server/tree/b8afdce9a3bc8fd927fe4ec9be2e910ee8f1635d/services/distribution/src/test/resources/parameters), [here](https://github.com/corona-warn-app/cwa-server/tree/d6c0ad58f6473bd55af9a0cbe2c0cc8db3914af6/services/distribution/src/main/resources/master-config) and possibly more).

## Additional context
- This "bug" doesn't have any technical implications and should only be fixed for consistency
- Downwards-compatibility of the proto messages should not be infringed by the proposed change

mibrasap:
## Checklist

## Description
Fix attenuation risk parameter naming to include "10 db"
closes #638


services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/appconfig/validation/ValidationError.java
```diff
@@ -80,6 +80,7 @@ public int hashCode() {
     INVALID_URL,
     INVALID_PARTITIONING,
     TOO_MANY_DECIMAL_PLACES,
-    MISSING_ENTRY
+    MISSING_ENTRY,
+    UNKOWN_ENTRY
```
pithumke, line 84: Is this error type actually being used somewhere?

services/distribution/src/test/java/app/coronawarn/server/services/distribution/assembly/appconfig/YamlLoaderTest.java
```diff
@@ -40,7 +40,8 @@ void okFile() throws UnableToLoadFileException {
   @ValueSource(strings = {
       "configtests/app-config_empty.yaml",
       "configtests/wrong_file.yaml",
-      "configtests/broken_syntax.yaml",
+      "configtests/app-config_broken_syntax.yaml",
+      "parameters/naming_mismatch.yaml",
```
pithumke, line 44: Would be nice to have the test files organized a bit more "cleanly". Looks like files belonging to similar test cases are scattered across multiple different directories.

services/distribution/src/test/java/app/coronawarn/server/services/distribution/assembly/appconfig/YamlLoaderTest.java
```diff
@@ -40,7 +40,8 @@ void okFile() throws UnableToLoadFileException {
   @ValueSource(strings = {
       "configtests/app-config_empty.yaml",
       "configtests/wrong_file.yaml",
-      "configtests/broken_syntax.yaml",
+      "configtests/app-config_broken_syntax.yaml",
```
pithumke, line 43: Was this renamed on purpose?

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/appconfig/validation/ValidationError.java
michael-burwig: Does not look like it.

services/distribution/src/test/java/app/coronawarn/server/services/distribution/assembly/appconfig/YamlLoaderTest.java
KevponSAP: Yes, because the file does not exist.

PR #647, opened 2020-07-02
downscale gracefully

### LINKED ISSUE #547: Failing requests in down-scaling scenario
## Describe the bug

Reducing the number of submission service service pods leads to immediate termination of request processing on the service instance that is shutting down while. This results in clients observing a failing requests.
In the worst case scenario, submission processing is interrupted after a TAN was verified successfully but the corresponding diagnosis keys have not been persisted yet.

## Expected behaviour

On a service instance that is shutting down, no new requests shall be accepted. Requests that are already being processed at the time shall be given a grace period in which they are allowed to finish processing.

## Possible Fix

- ``ThreadPoolExecutor.awaitTermination`` and similar concepts

## Additional context
From a previous performance testing run:

image

hilmarf:
closes #547

## Description

When the JVM receives the SIGTERM, shutdown hooks are being executed. Unfortunately there isn't one automatically existing for Tomcat. Because of async processing (use of [DeferredResult](https://github.com/corona-warn-app/cwa-server/blob/v1.0.9/services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java#L95)), there may be still some open requests, that Tomcat has to respond on. So with this PR, the Tomcat connector is paused (it's not taking new requests) and the ThreadPoolExecutor get's 30 seconds to finish its work before it's terminated.


services/submission/src/main/java/app/coronawarn/server/services/submission/shutdown/WaitOnTomcatToProcessPendingRequests.java
```diff
@@ -0,0 +1,64 @@
...
+public class WaitOnTomcatToProcessPendingRequests
+    implements TomcatConnectorCustomizer, ApplicationListener<ContextClosedEvent> {
+
+  private static final Logger logger = LoggerFactory.getLogger(WaitOnTomcatToProcessPendingRequests.class);
+
+  private Connector connector;
+
+  @Override
+  public void onApplicationEvent(ContextClosedEvent event) {
+    logger.info("receiced ContextClosedEvent '{}', pausing the connector...", event);
```
johanneseschrig, line 42:
```suggestion
    logger.info("received ContextClosedEvent '{}', pausing the connector...", event);
```

services/submission/src/main/java/app/coronawarn/server/services/submission/shutdown/WaitOnTomcatToProcessPendingRequests.java
```diff
@@ -0,0 +1,64 @@
...
+    this.connector.pause();
+    Executor executor = this.connector.getProtocolHandler().getExecutor();
+    if (executor instanceof ThreadPoolExecutor) {
+      try {
+        ThreadPoolExecutor threadPoolExecutor = (ThreadPoolExecutor) executor;
+        threadPoolExecutor.shutdown();
+        if (!threadPoolExecutor.awaitTermination(30, TimeUnit.SECONDS)) {
```
pithumke, line 49: Should probably be configurable through the `SubmissionServiceConfiguration`.

services/submission/src/main/java/app/coronawarn/server/services/submission/shutdown/WaitOnTomcatToProcessPendingRequests.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import org.apache.catalina.connector.Connector;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.boot.web.embedded.tomcat.TomcatConnectorCustomizer;
+import org.springframework.context.ApplicationListener;
+import org.springframework.context.event.ContextClosedEvent;
+
+public class WaitOnTomcatToProcessPendingRequests
```
pithumke, line 33: Just a personal preference, but I would opt for a class name like `GracefulShutdownConnectorCustomizer` or `GracefulShutdownApplicationListener`. The current class name reads like a sentence - something I would expect for a method name. For class names I personally prefer something that reads more like an entity name.

services/submission/src/main/java/app/coronawarn/server/services/submission/shutdown/WaitOnTomcatToProcessPendingRequests.java
```diff
@@ -0,0 +1,64 @@
...
+    implements TomcatConnectorCustomizer, ApplicationListener<ContextClosedEvent> {
+
+  private static final Logger logger = LoggerFactory.getLogger(WaitOnTomcatToProcessPendingRequests.class);
+
+  private Connector connector;
+
+  @Override
+  public void onApplicationEvent(ContextClosedEvent event) {
+    logger.info("receiced ContextClosedEvent '{}', pausing the connector...", event);
+    this.connector.pause();
```
pithumke, line 43: Could this throw a `NullPointerException`?

services/submission/src/main/java/app/coronawarn/server/services/submission/shutdown/WaitOnTomcatToProcessPendingRequests.java
pithumke, line 33: A few lines of JavaDoc would be appreciated.

services/submission/src/main/java/app/coronawarn/server/services/submission/ServerApplication.java
```diff
@@ -59,6 +64,26 @@ TimedAspect timedAspect(MeterRegistry registry) {
...
+  @Bean
+  public ConfigurableServletWebServerFactory webServerFactory(WaitOnTomcatToProcessPendingRequests gracefulShutdown) {
+    TomcatServletWebServerFactory factory = new TomcatServletWebServerFactory();
+    factory.addConnectorCustomizers(gracefulShutdown);
+    return factory;
+  }
+
+  @PreDestroy
+  public void onDestroy() {
+    LogManager.shutdown();
```
pithumke, line 84: If we shut down the LogManager on "pre-destroy", does this mean that we can not log stuff anymore *during* destroy?

christian-kirschnick: I believe we should have a consistent implementation for this - the distribution service has the same functionality, but not using `javax.*` but instead spring boot style with `DisposableBean`. Let's do the same here.

https://github.com/corona-warn-app/cwa-server/blob/f41fcbad05acf1e0581dc7a4c8bac5cf76567620/services/distribution/src/main/java/app/coronawarn/server/services/distribution/Application.java#L58-L65

michael-burwig: Not an answer to the question, but I assume this relates to the same underlying issue as in [#589](https://github.com/corona-warn-app/cwa-server/issues/589).

hilmarf: available [shutdown-callbacks](https://www.baeldung.com/spring-shutdown-callbacks)

Destroy methods are called in the order:
1. Methods annotated with @PreDestroy
1. destroy() as defined by the DisposableBean callback interface
1. A custom configured destroy() method

So which approach, has to be followed, when there are multiple ways of doing things?

Side note @pithumke: distribution service 
[logger.error("Application terminated abnormally.");](https://github.com/corona-warn-app/cwa-server/blob/master/services/distribution/src/main/java/app/coronawarn/server/services/distribution/Application.java#L72)
I guess this won't work, because `SpringApplication.exit(appContext);` earlier triggers the destroy method, which shuts down the logmanager...

michael-burwig: - `@PreDestroy` is somewhat deprecated with Java 11 (not sure which dependency makes it available in the submission module right now).
- Distribution module already uses Spring's `DisposableBean` interface as a non-deprecated way of achieving the desired functionality.

hilmarf: now using `DisposableBean`

PR #651, opened 2020-07-05
Make .env config advice more explicit

llewelld:
## Checklist

* **N/A** Test your changes as thoroughly as possible before you commit them. Preferably, automate your test by unit/integration tests.
* **N/A** Make sure `mvn install` runs for the whole project and, if you touched any code in the respective service, submission and distribution service can be run with `spring-boot:run`

N/A since the changes only affect the README.md.

## Description

The servers will only run correctly if the .env file has sensible values. This change to the ReadMe makes clearer what's needed when running using Docker Compose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ If you want to use Docker-based deployment, you need to install Docker on your l
 
 #### Running the Full CWA Backend Using Docker Compose
 
-For your convenience, a full setup including the generation of test data has been prepared using [Docker Compose](https://docs.docker.com/compose/reference/overview/). To build the backend services, run ```docker-compose build``` in the repository's root directory. A default configuration file can be found under ```.env``` in the root folder of the repository. The default values for the local Postgres and Zenko Cloudserver should be changed in this file before docker-compose is run.
+For your convenience, a full setup for local development and testing purposes, including the generation of test data has been prepared using [Docker Compose](https://docs.docker.com/compose/reference/overview/). To build the backend services, run ```docker-compose build``` in the repository's root directory. A default configuration file can be found under ```.env``` in the root folder of the repository. If the endpoints are to be exposed to the network the default values in this file should be changed before docker-compose is run.
 
 Once the services are built, you can start the whole backend using ```docker-compose up```.
 The distribution service runs once and then finishes. If you want to trigger additional distribution runs, run ```docker-compose run distribution```.


README.md
```diff
@@ -49,7 +49,7 @@ If you want to use Docker-based deployment, you need to install Docker on your l
...
+For your convenience, a full setup including the generation of test data has been prepared using [Docker Compose](https://docs.docker.com/compose/reference/overview/). To build the backend services, run ```docker-compose build``` in the repository's root directory. A default configuration file can be found under ```.env``` in the root folder of the repository. The default values should be changed in this file before docker-compose is run. In particular, choose a more secure password for the Postgres server, then change the username and password values for **DISTRIBUTION**, **SUBMISSION** and **FLYWAY** to match the Postgres values. The **PGADMIN** password should also be changed, but needn't match the Postgres password.
```
stevesap, line 52:
```suggestion
For your convenience, a full setup for local development and testing purposes, including the generation of test data has been prepared using [Docker Compose](https://docs.docker.com/compose/reference/overview/). To build the backend services, run ```docker-compose build``` in the repository's root directory. A default configuration file can be found under ```.env``` in the root folder of the repository. The default values may be changed in this file before docker-compose is run.
```

llewelld: The clarity that there's no need to change the values for a local install is helpful, but perhaps it's also useful to retain the info about which values must match in order for the system to work. For example, in my situation it's not viable for me to run the server locally as I need to test with a server on the public/cellular network. In this case understanding what changes are needed to the configuration is useful.

stevesap: The passwords for the distribution, submission and flyway users do not need to match the Postgres user one as in fact, by default they do not (see the relevant PR and introduction of users to local setup at https://github.com/corona-warn-app/cwa-server/pull/552/files#diff-e85991ccbb191b6af3f4542d855b5d24 for more details). Also, the particular instructions here refer strictly to the scenario where someone wants to set up docker compose for a local dev/test setup, rather than the scenario you are describing.

llewelld: Thanks for the link; that certainly does seem to make the point about them needing to be the same redundant.

To clarify concerning the local/not local setup, what I mean is that the endpoint services are exposed to the network, not that they're not necessarily running locally. My reading of the original documentation was that this situation was covered (it talks about local deployment, but not about what should/shouldn't be network-accessible). In this case, as long as Posgres and PgAdmin aren't exposed, but the other endpoints are, is it safe to leave the .env file as-is? I ask because the current instructions imply otherwise.

stevesap: I think your scenario is not a common one as most devs would just use the docker compose locally and privately, but in your case if you plan to expose the services publicly, then probably it would be a good idea to secure them further with stronger passwords. However, I believe the suggestion "The default values may be changed in this file before docker-compose is run" is sufficient in implying that this is an option for anyone who needs to do something like that for their particular scenario.

llewelld: Honestly I have no idea what scenarios would be more or less common, but my scenario does at least seem like a reasonable usecase to me. I'm very happy with your changes removing the incorrect point about values needing to be the same, but the final sentence shifts the balance in the text from "you should change the values to make them more secure" to "don't expose this to the network", which is the opposite of what I was aiming for here. It would make more sense if it said something like "If the endpoints are to be exposed to the network the default values in this file should be changed before docker-compose is run."

llewelld: Pushed an update based on the discussion so far.

PR #653, opened 2020-07-06
Addition to CODEOWNERS.md

EugenM-SAP:
## Checklist


lotharschulz:
[CODEOWNERS syntax](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax) supports teams via `@org/team-name` syntax. This way you can reference team(s) and edit team(s) to potentially safe PRs like this.

EugenM-SAP:
@lotharschulz  thanks for your suggestion, however the consensus is that code owners should be explicitly stated in the markdown file, such that modifications to this list are transparent and subject to pull requests. Also, the teams currently are  setup in such a way that they include a larger set of  people than what should be in the codeowners.md.

PR #659, opened 2020-07-10
Feature/padding

### LINKED ISSUE #658: Plausible Deniability: Add "Padding" property to submission payload
Currently, mobile apps are collecting keys based on their install date. When a user uploads his keys (after being tested positive), the payload will contain 1-13 keys. Although the traffic between mobile & server is secured, an attacker may still sniff the packages in the network and predict, based on the request size, how many keys are probably part of the submission request. This may then lead to additional information for the attacker in an attempt to deanonymize a user.

In order to mitigate this attack, add a new field to the submission payload, called `padding` of type string. The field shall be optional and must not break compatibility to the current/older mobile implementations (which should be given due to the use of protobuf). The server shall not process this field.

mibrasap:
## Checklist

## Description

added new string "padding" in Submission Payload to enable plausible deniability
closes #658

### DIFF
--- a/common/protocols/src/main/proto/app/coronawarn/server/common/protocols/internal/submission_payload.proto
+++ b/common/protocols/src/main/proto/app/coronawarn/server/common/protocols/internal/submission_payload.proto
@@ -6,5 +6,5 @@ import "app/coronawarn/server/common/protocols/external/exposurenotification/tem
 
 message SubmissionPayload {
   repeated app.coronawarn.server.common.protocols.external.exposurenotification.TemporaryExposureKey keys = 1;
-  string padding = 2;
+  bytes padding = 2;
 }
--- a/common/protocols/src/main/proto/app/coronawarn/server/common/protocols/internal/submission_payload.proto
+++ b/common/protocols/src/main/proto/app/coronawarn/server/common/protocols/internal/submission_payload.proto
@@ -6,4 +6,5 @@ import "app/coronawarn/server/common/protocols/external/exposurenotification/tem
 
 message SubmissionPayload {
   repeated app.coronawarn.server.common.protocols.external.exposurenotification.TemporaryExposureKey keys = 1;
+  string padding = 2;
 }
--- a/services/submission/src/main/java/app/coronawarn/server/services/submission/config/SubmissionPayloadSizeFilter.java
+++ b/services/submission/src/main/java/app/coronawarn/server/services/submission/config/SubmissionPayloadSizeFilter.java
@@ -12,7 +12,7 @@ import org.springframework.web.filter.OncePerRequestFilter;
 @Component
 public class SubmissionPayloadSizeFilter extends OncePerRequestFilter {
 
-  private static final int MAX_REQUEST_SIZE = 100 * 1024;
+  public static final int MAX_REQUEST_SIZE = 100 * 1024;
 
   /**
    * Filters each request that exceeds the maximum size of 100KB
--- a/services/submission/src/main/java/app/coronawarn/server/services/submission/config/SubmissionPayloadSizeFilter.java
+++ b/services/submission/src/main/java/app/coronawarn/server/services/submission/config/SubmissionPayloadSizeFilter.java
@@ -35,13 +35,7 @@ public class SubmissionPayloadSizeFilter extends OncePerRequestFilter {
   public static final int MAX_REQUEST_SIZE = 100 * 1024;
 
   /**
-   * Filters each request that exceeds the maximum size of 100KB
-   *
-   * @param request
-   * @param response
-   * @param filterChain
-   * @throws ServletException
-   * @throws IOException
+   * Filters each request that exceeds the maximum size of 100KB.
    */
   @Override
   protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
--- a/services/submission/src/main/java/app/coronawarn/server/services/submission/config/SubmissionPayloadSizeFilter.java
+++ b/services/submission/src/main/java/app/coronawarn/server/services/submission/config/SubmissionPayloadSizeFilter.java
@@ -1,0 +1,37 @@
+package app.coronawarn.server.services.submission.config;
+
+import java.io.IOException;
+import javax.servlet.FilterChain;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.springframework.http.HttpStatus;
+import org.springframework.stereotype.Component;
+import org.springframework.web.filter.OncePerRequestFilter;
+
+@Component
+public class SubmissionPayloadSizeFilter extends OncePerRequestFilter {
+
+  private static final int MAX_REQUEST_SIZE = 100 * 1024;
+
+  /**
+   * Filters each request that exceeds the maximum size of 100KB
+   *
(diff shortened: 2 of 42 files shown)


services/submission/src/test/java/app/coronawarn/server/services/submission/controller/RequestExecutor.java
```diff
@@ -62,6 +62,10 @@ public RequestExecutor(TestRestTemplate testRestTemplate) {
 
   public ResponseEntity<Void> executePost(Collection<TemporaryExposureKey> keys, HttpHeaders headers) {
     SubmissionPayload body = SubmissionPayload.newBuilder().addAllKeys(keys).build();
+    return executePost(body, headers);
```
pithumke, line 65: Is the "old" method with the `Collection<TemporaryExposureKey> keys` parameter still needed?

common/protocols/src/main/proto/app/coronawarn/server/common/protocols/internal/submission_payload.proto
```diff
@@ -6,4 +6,5 @@ import "app/coronawarn/server/common/protocols/external/exposurenotification/tem
 
 message SubmissionPayload {
   repeated app.coronawarn.server.common.protocols.external.exposurenotification.TemporaryExposureKey keys = 1;
-}
\ No newline at end of file
+  string padding = 2;
```
pithumke, line 9: Should we maybe have a size constraint for this somewhere?

pithumke: Consider padding the HTTP package with something like a `cwa-padding` header. This wouldn't require any changes on the server side and could be implemented 100% on the mobile client side.

pithumke: If we put the padding into the proto, consider `byte[]` instead of `string` (reduced parsing overhead for data that is thrown away anyways).

christian-kirschnick: Let's keep the proto change as it is automatically downwards compatible and more explicit.

In terms of byte vs string: I guess you are right on the receiving side - should also be OK on the sender side. So I'd be fine with it.

For the length restriction: I don't think that this is available in standard protobuf - I think we should just restrict the maximum body size for the tomcat configuration - that way the potential attack will be rejected already at webserver level, and does not hit Spring/Controller/protobuf deserialization.

services/submission/src/main/java/app/coronawarn/server/services/submission/config/SubmissionPayloadSizeFilter.java
```diff
@@ -0,0 +1,51 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.springframework.http.HttpStatus;
+import org.springframework.stereotype.Component;
+import org.springframework.web.filter.OncePerRequestFilter;
+
+@Component
+public class SubmissionPayloadSizeFilter extends OncePerRequestFilter {
+
+  public static final int MAX_REQUEST_SIZE = 100 * 1024;
```
stevesap, line 35: Consider externalizing this value to application configuration so it can be easily changed on production if needed

services/submission/src/main/java/app/coronawarn/server/services/submission/config/SubmissionPayloadSizeFilter.java
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * Filters each request that exceeds the maximum size of 100KB.
+   */
+  @Override
+  protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
+      throws ServletException, IOException {
+    if (request.getContentLengthLong() > MAX_REQUEST_SIZE) {
```
christian-kirschnick, line 43: Did you try setting `server.tomcat.max-http-form-post-size=100KB` in the `application.yaml`? It should also work.

services/submission/src/test/java/app/coronawarn/server/services/submission/controller/RequestExecutor.java
mibrasap: Good hint, refactored

common/protocols/src/main/proto/app/coronawarn/server/common/protocols/internal/submission_payload.proto
mibrasap: As discussed, we limit the submission request size to 100KB

services/submission/src/main/java/app/coronawarn/server/services/submission/config/SubmissionPayloadSizeFilter.java
mibrasap: Good hint, refactored

services/submission/src/main/java/app/coronawarn/server/services/submission/config/SubmissionPayloadSizeFilter.java
```diff
@@ -0,0 +1,55 @@
...
+public class SubmissionPayloadSizeFilter extends OncePerRequestFilter {
+
+  private final Integer maximumRequestSize;
+
+  public SubmissionPayloadSizeFilter(SubmissionServiceConfig config) {
+    this.maximumRequestSize = config.getMaximumRequestSize();
+  }
+
+  /**
+   * Filters each request that exceeds the maximum size of 100KB.
```
MKusber, line 42: Please update the comment to reflect the change to an externalized property

services/submission/src/main/resources/application.yaml
```diff
@@ -23,6 +23,8 @@ services:
     random-key-padding-multiplier: ${RANDOM_KEY_PADDING_MULTIPLIER:1}
     # The ApacheHttpClient's connection pool size.
     connection-pool-size: 200
+    # The maximum request size accepted by the SubmissionController in Bytes.
+    maximum-request-size: 102400
```
MKusber, line 27: According to @stevesap 's comment, I would ask to consider having this an environment variable with default. Please see e.g. `random-key-paddin-multiplier`.

services/submission/src/main/java/app/coronawarn/server/services/submission/config/SubmissionServiceConfig.java
```diff
@@ -86,6 +87,14 @@ public void setConnectionPoolSize(Integer connectionPoolSize) {
     this.connectionPoolSize = connectionPoolSize;
   }
 
+  public Integer getMaximumRequestSize() {
```
MKusber, line 90: Please consider setting some limits, like for the random key multiplier.

services/submission/src/main/resources/application.yaml
```diff
@@ -23,6 +23,8 @@ services:
     random-key-padding-multiplier: ${RANDOM_KEY_PADDING_MULTIPLIER:1}
     # The ApacheHttpClient's connection pool size.
     connection-pool-size: 200
+    # The maximum request size accepted by the SubmissionController in Bytes.
```
christian-kirschnick, line 26: Adjust comment here -> should probably suggest examples (e.g. 200B or 100KB)

PR #664, opened 2020-07-13
Enhancement/time zone handling

sorin-stefan-iovita-sap:
### Description:

- Call LocalDateTime.now(ZoneOffset.UTC) once at distribution start and pass it through the code base.

- Add unit test that uses current hour as distribution time to prove time zone handling.

- Generate test keys up to current hour

### DIFF
--- a/services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/structure/util/TimeUtils.java
+++ b/services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/structure/util/TimeUtils.java
@@ -28,9 +28,11 @@ import java.time.temporal.ChronoUnit;
 
 public class TimeUtils {
 
-  private static LocalDate utcDate;
   private static LocalDateTime utcHour;
   private static Instant now;
+
+  private TimeUtils() {
+  }
 
   /**
    * Returns the UTC date and time at the beginning of the current hour or creates a new instance if called the first
@@ -45,8 +47,9 @@ public class TimeUtils {
 
 
   /**
-   * Injects UTC date time value
-   * @param dateTime
+   * Injects UTC date time value.
+   *
+   * @param dateTime a LocalDateTime object in UTC timezone.
    */
   public static void setUtcHour(LocalDateTime dateTime) {
     utcHour = dateTime;
@@ -56,18 +59,7 @@ public class TimeUtils {
    * Returns the UTC date or creates a new instance if called the first time.
    */
   public static LocalDate getUtcDate() {
-    if (utcDate == null) {
-      utcDate = LocalDate.now(ZoneOffset.UTC);
-    }
-    return utcDate;
-  }
-
-  /**
-   * Injects UTC date value
-   * @param date
-   */
-  public static void setUtcDate(LocalDate date) {
-    utcDate = date;
+    return getCurrentUtcHour().toLocalDate();
   }
 
   /**
--- a/services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/structure/util/TimeUtils.java
+++ b/services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/structure/util/TimeUtils.java
@@ -1,9 +1,30 @@
+/*-
+ * ---license-start
+ * Corona-Warn-App
+ * ---
+ * Copyright (C) 2020 SAP SE and all other contributors
+ * ---
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ---license-end
+ */
+
 package app.coronawarn.server.services.distribution.assembly.structure.util;
 
 import java.time.Instant;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.ZoneOffset;
+import java.time.temporal.ChronoUnit;
 
 public class TimeUtils {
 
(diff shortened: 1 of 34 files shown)


fredrb:
Once the application was using UTC for all dates, we realised that test data generation was generating keys up to current hour - 2. 

We changed the behavior so that keys are generated up to current hour always. The assembly step will then select keys up until the previous hour or current hour if demo step is active. Without this change, running test generation with or without demo profile didn't make any difference in the output folder.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/structure/util/TimeUtils.java
```diff
@@ -0,0 +1,65 @@
...
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+import java.time.temporal.ChronoUnit;
+
+public class TimeUtils {
+
+  private static LocalDate utcDate;
+  private static LocalDateTime utcHour;
+  private static Instant now;
```
michael-burwig, line 33: Is there a reason for having these as singletons that are independent of each other?
You could save a couple LoCs by:
- exposing these fields as `public`, `final` constants
- initializing them right here
- adjust the names and javadoc

What do you think?

sorin-stefan-iovita-sap: I was thinking that these fields can be changed by adding setters (in the future) so it can be easier to test certain components which uses them.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/structure/util/TimeUtils.java
```diff
@@ -0,0 +1,65 @@
...
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+import java.time.temporal.ChronoUnit;
+
+public class TimeUtils {
+
+  private static LocalDate utcDate;
+  private static LocalDateTime utcHour;
```
fredrb, line 32: Maybe we should call this `utcTime` other than `utcHour`?

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/structure/util/TimeUtils.java
fredrb: Also, shouldn't we be using one LocalDate/LocalDateTime object and then converting to other formats if necessary? 

With the current solution it could happen that if `getCurrentUtcHour()` is called for the first time at 18:59 and `getUtcDate()` is called for the first time at 19:00, we have different hour values for these two properties. 

This would be solved with Michael's suggestion of keeping the static fields public because then the three values would be instantiated virtually together. But then mocking becomes a bit redundant. 

To inject values in this class on a test I had to do the following:
```java
TimeUtils.setUtcDate(LocalDate.of(1970, 1, 5));
TimeUtils.setUtcHour(LocalDateTime.of(1970, 1, 5, 0, 0));
```

Ideally, only the second line should be sufficient, and when the code call `getUtcDate()` it should return `5.1.1970`.

So my suggestion is that we keep only `LocalDateTime` property and implement `getUtcDate` something like this:
```java
public static LocalDate getUtcDate() {
  return TimeUtils.getCurrentUtcHour().toLocalDate()
}
```

Unless we have a scenario where want to inject different values?

sorin-stefan-iovita-sap: Done. 😄

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/structure/util/TimeUtils.java
sorin-stefan-iovita-sap: utcTime would be missleading since this is the time truncated to hour. i.e. 2020-07-14T10:00.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/structure/util/TimeUtils.java
```diff
@@ -0,0 +1,74 @@
...
+  private TimeUtils() {
+  }
+
+  /**
+   * Returns the UTC date and time at the beginning of the current hour or creates a new instance if called the first
+   * time.
+   */
+  public static LocalDateTime getCurrentUtcHour() {
+    if (utcHour == null) {
+      utcHour = LocalDateTime.now(ZoneOffset.UTC).truncatedTo(ChronoUnit.HOURS);
```
michael-burwig, line 43: Would this be equivalent to `utcHour = LocalDateTime.from(getNow().truncatedTo(ChronoUnit.HOURS));`?

If so, is there a reason to initialize the fields independent of each other?

sorin-stefan-iovita-sap: Yes, that would be the equivalent. 👍 
There is no reason to initialize the fields independent of each other.
 I'll do the changes.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/structure/util/TimeUtils.java
```diff
@@ -0,0 +1,68 @@
...
+   * Returns the UTC date.
+   */
+  public static LocalDate getUtcDate() {
+    return getCurrentUtcHour().toLocalDate();
+  }
+
+  /**
+   * Returns the UTC {@link Instant} time or creates a new instance if called the first time.
+   */
+  public static Instant getNow() {
```
michael-burwig, line 53: I find the terms `now` and `current` used in a misleading manner here, since what you get is rather _"At some point in the past OR now"_.
But if that's just me, I do not mind.

In general, dependency injecting a "startupTime" bean into whichever component needs it, seems like the cleaner solution to me. That said, I see how this is not straightforward right now since many of the places that make use of this timestamp do not use dependency injection. So all in all, I consider your solution is acceptable.

sorin-stefan-iovita-sap: Initially I wanted to implement a bean and spread it across the application using dependency injection but that could not be done because many classes where it could be used are not components.

PR #668, opened 2020-07-15
Feature/semantic config validation

UnchartedBull:
Fixed the URL with Port regex to also accept hyphens. Worked on @MKusber's machine as well, should be deployed with caution though ...


services/submission/src/main/java/app/coronawarn/server/services/submission/config/SubmissionServiceConfig.java
```diff
@@ -32,14 +33,25 @@
 @Validated
 public class SubmissionServiceConfig {
 
+  private static final String PATH_REGEX = "^[/]?[a-zA-Z0-9_]+[/[a-zA-Z0-9_]+]*$";
+  private static final String URL_WITH_PORT_REGEX = "^http[s]?://[a-z0-9-]+(\\.[a-z0-9-]+)*:[0-9]{2,6}";
```
stevesap, line 37:
```suggestion
  private static final String URL_WITH_PORT_REGEX = "^http[s]?://[a-z0-9-]+(\\.[a-z0-9-]+)*:[0-9]{2,6}$";
```

PR #686, opened 2020-07-28
Usage of postgres in test environment for database consistency

ioangut:
## Description

* Replaced h2 database for testing with postgresql
* Removed vendor specific flyway migration in testing


fredrb:
For reference: #533

PR #687, opened 2020-07-29
Feature/federation distribution service

fredrb:
**(!) This is an intermediate PR for the current federation gateway changes**
This PR adds the following changes:
1. `origin_country` and `visited_countries` column to DiagnosisKey table, together with migration scripts
2. Replaces H2 with Postgres test containers (already covered by PR #686 and merged to master)
3. Consider a new environment variable `DISTRIBUTION_COUNTRY` on the Distribution Service run. This will be propagated to the configuration bean and used across the service
4. Generate Test Data with `testdata` profile with `origin_country` and `visited_countries` information. Currently it will generate data for country received as `DISTRIBUTION_COUNTRY` with a random boolean if it's a traveler or not.
5. Enhance retention policy to consider `DISTRIBUTION_COUNTRY` and only delete keys from DB that contain the given country
6. Quite a few enhancements in the `persistence` package to support the two new values. 

A few things to keep in mind:
1. If you already had a working version of the server locally, would be good to delete the container **and volume** before running the migrations. There are new migrations but also new roles/users. If you don't want to delete the container completely, you can run the following statement in your instance:
```sql
CREATE ROLE cwa_federation_callback;
CREATE USER "local_setup_callback" WITH INHERIT IN ROLE cwa_federation_callback ENCRYPTED PASSWORD 'local_setup_callback';
```
2. The build is breaking for the `callback` service, which I think it's because of the merge between the downloader branch and this one. It seems to be because the new service still has configuration to use H2 other than Postgres for testing. @ioangut and @KevponSAP could you guys check?


common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/DiagnosisKey.java
```diff
@@ -126,6 +136,22 @@ public long getSubmissionTimestamp() {
     return submissionTimestamp;
   }
 
+  public String getOriginCountry() {
+    return originCountry;
+  }
+
+  public void setOriginCountry(String originCountry) {
```
EugenM-SAP, line 143: I believe we can remove the setters to preserve imutability

common/persistence/src/main/java/app/coronawarn/server/common/persistence/repository/DiagnosisKeyRepository.java
```diff
@@ -39,6 +39,27 @@
   @Query("SELECT COUNT(*) FROM diagnosis_key WHERE submission_timestamp<=:threshold")
   int countOlderThanOrEqual(@Param("threshold") long submissionTimestamp);
 
+  /**
+   * Counts all entries that have a submission timestamp less or equal than the specified one
+   *  and match the origin country_code.
```
EugenM-SAP, line 44: was the origin word  intended in the doc , or is the meaning more like "and match the given country_code"

common/persistence/src/main/java/app/coronawarn/server/common/persistence/repository/DiagnosisKeyRepository.java
```diff
@@ -69,4 +100,20 @@ void saveDoNothingOnConflict(
...
       @Param("transmissionRisk") int transmissionRisk);
+
+  @Modifying
+  @Query("INSERT INTO diagnosis_key "
+      + "(key_data, rolling_start_interval_number, rolling_period, submission_timestamp, transmission_risk_level, "
+        + "origin_country, visited_countries) "
+      + "VALUES (:keyData, :rollingStartIntervalNumber, :rollingPeriod, :submissionTimestamp, :transmissionRisk, "
+        + ":origin_country, :visited_countries) "
+      + "ON CONFLICT DO NOTHING")
+  void saveDoNothingOnConflictCountries(
```
EugenM-SAP, line 111: Might be a good thing  to completely replace the old saveDoNothingOnConflict with the new insert statement that includes countries

common/persistence/src/main/java/app/coronawarn/server/common/persistence/service/DiagnosisKeyService.java
```diff
@@ -61,9 +63,28 @@ public DiagnosisKeyService(DiagnosisKeyRepository keyRepository) {
...
+  /**
+   * Persists the specified collection of {@link DiagnosisKey} instances. If the key data of a particular diagnosis key
+   * already exists in the database, this diagnosis key is not persisted.
+   *
+   * @param diagnosisKeys must not contain {@literal null}.
+   * @throws IllegalArgumentException in case the given collection contains {@literal null}.
+   */
+  @Timed
+  @Transactional
+  public void saveDiagnosisKeysWithCountry(Collection<DiagnosisKey> diagnosisKeys) {
```
EugenM-SAP, line 82: Is the special variant needed or could we replace the existing save method compeltely?

common/persistence/src/main/java/app/coronawarn/server/common/persistence/service/DiagnosisKeyService.java
```diff
@@ -113,9 +151,13 @@ public void applyRetentionPolicy(int daysToRetain) {
...
         .toEpochSecond(UTC) / SECONDS_PER_HOUR;
-    int numberOfDeletions = keyRepository.countOlderThanOrEqual(threshold);
+    int numberOfDeletions = keyRepository.countOlderThanOrEqual(threshold, countryCode);
     logger.info("Deleting {} diagnosis key(s) with a submission timestamp older than {} day(s) ago.",
         numberOfDeletions, daysToRetain);
-    keyRepository.deleteOlderThanOrEqual(threshold);
+    keyRepository.deleteOlderThanOrEqual(threshold, countryCode);
+  }
+
+  public void applyRetentionPolicy(int daysToRetain) {
```
EugenM-SAP, line 160: IT might cause problems if we leave this method as part of the DiagnosisKeyService API, because clients don't know what the behaviour is and which would be the default country. I think it's better if we remove this code path and enforce callers to specify the country. Or we could move the DEFAULT_COUNTRY one level higher and provide it to the service at construction time (as an env variable)

services/distribution/src/main/java/app/coronawarn/server/services/distribution/runner/TestDataGeneration.java
```diff
@@ -156,15 +164,30 @@ private long getRetentionStartTimestamp() {
...
   }
 
+  /**
+   * Either returns the list of all possible visited countries or only current distribution
+   * This ensure that when test generation runs for a country (e.g. FR) all keys in the
+   * distribution will contain FR in the vistied_countries array.
+   */
+  private List<String> generateListOfVisitedCountries(String distributionCountry) {
+    if (random.nextBoolean()) {
+      return List.of("DE", "DK", "FR");
```
EugenM-SAP, line 174: based on the doc , the distributionCountry param should be part of the list right?

common/persistence/src/main/java/app/coronawarn/server/common/persistence/repository/DiagnosisKeyRepository.java
fredrb: The meaning was supposed to be something like "match the given country_code", yes. I will rephrase this.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/runner/TestDataGeneration.java
fredrb: Yes, it should. But this is as a placeholder for the future "list of all countries". Since all the tests use either "DE", "DK" or "FR" I left it as this for now. 

Maybe we should add a new configuration in the yaml file that will be this "list of all countries"? A static list of country keys that will be used here and any other places where such list is needed. 

The whole point is that we need to get this list of valid countries from somewhere.

common/persistence/src/main/java/app/coronawarn/server/common/persistence/service/DiagnosisKeyService.java
```diff
@@ -104,7 +142,7 @@ private static boolean isDiagnosisKeyValid(DiagnosisKey diagnosisKey) {
    * @throws IllegalArgumentException if {@code daysToRetain} is negative.
```
MichaelPFrey, line 142:
```suggestion
   * @param countryCode country filter.
   * @throws IllegalArgumentException if {@code daysToRetain} is negative.
```

common/persistence/src/main/java/app/coronawarn/server/common/persistence/service/DiagnosisKeyService.java
```diff
@@ -113,9 +151,13 @@ public void applyRetentionPolicy(int daysToRetain) {
         .ofInstant(Instant.now(), UTC)
         .minusDays(daysToRetain)
         .toEpochSecond(UTC) / SECONDS_PER_HOUR;
-    int numberOfDeletions = keyRepository.countOlderThanOrEqual(threshold);
+    int numberOfDeletions = keyRepository.countOlderThanOrEqual(threshold, countryCode);
     logger.info("Deleting {} diagnosis key(s) with a submission timestamp older than {} day(s) ago.",
         numberOfDeletions, daysToRetain);
```
MichaelPFrey, line 156: why does the logger not record the countryCode?
Can the countryCode be reliably reconstructed by looking at log infos before/after this log line?

fredrb: Good point. Ideally we should enhance all logs with a country prefix.

michael-burwig:
What about the 14 sonar code smells?

services/distribution/src/main/java/app/coronawarn/server/services/distribution/runner/TestDataGeneration.java
EugenM-SAP: I agree we should have environment variable for that, but i guess it can be done outside this PR

PR #691, opened 2020-07-31
Feature/federation submission validation

EugenM-SAP:
## Checklist

## Description

Partial implementation of changes for backlog item "Validate payload for Submission Service" with the following:
* Environment variable which holds allowed submission countries defined in the submission/application.yaml
* Include origin country in the submission payload validation as mandatory field (return 400 if not valid)
** Empty, null or a value which is not part of the allowed-countries list will result in a 400 http error
* Include visited countries in the submission payload validation as optional field. In case this field is invalid
* Commented unit tests waiting for the full payload proto definition

As there is a dependency on a separate backlog item which deals with the submission payload proto definition, there 
will be a follow-up PR which uncomments the tests and wraps up the validation logic

### DIFF
--- a/services/submission/src/main/java/app/coronawarn/server/services/submission/config/SubmissionServiceConfig.java
+++ b/services/submission/src/main/java/app/coronawarn/server/services/submission/config/SubmissionServiceConfig.java
@@ -23,7 +23,6 @@ package app.coronawarn.server.services.submission.config;
 import java.io.File;
 import java.util.Arrays;
 import java.util.List;
-
 import javax.validation.constraints.Max;
 import javax.validation.constraints.Min;
 import javax.validation.constraints.Pattern;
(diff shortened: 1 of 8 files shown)


services/submission/src/main/java/app/coronawarn/server/services/submission/config/SubmissionServiceConfig.java
```diff
@@ -36,6 +38,7 @@
 
   private static final String PATH_REGEX = "^[/]?[a-zA-Z0-9_]+[/[a-zA-Z0-9_]+]*$";
   private static final String URL_WITH_PORT_REGEX = "^http[s]?://[a-z0-9-]+(\\.[a-z0-9-]+)*(:[0-9]{2,6})?$";
+  private static final String ALLOWED_COUNTRY_CODES_REGEX = "^([a-zA-Z]{2}(\\,*[a-zA-Z]{2})*)$";
```
MichaelPFrey, line 41: Looking at `[a-zA-Z]{2}`:

When you allow lower and uppercase, all country code compare functions need to be case insensitive.

When the country code compare function need to be case insensitive, then there need to be test cases that ensure/test case insensitive.

Off-course: only allowing either upper or lower case also requires test case - but at least to me input sanitizing sounds easier then tracing down issues with different environments, inputs, code execution paths, string compares (both Java and SQL) etc.

EugenM-SAP: You make a good point about the potential issues that could appear with this kind of relaxation of constraints around the input.
I've also given thought to it during implementation, but because the ISO specification doesn't enforce lower/upper cases for these country codes, i decided to make this startup input a little more robust for any scripts or infrastructure that provides them (plan is to expose as app parameter not just statically defined in the yaml)
The check/logic done against this variable is case ignorant (see SubmissionServiceConfig.isCountryCodeAllowed), however, you're right, and i do plan to  add more test cases to cover this.
I will analyze again the potential risks though.

PR #710, opened 2020-08-14
Federation-Distribution Country

sorin-stefan-iovita-sap:
### Description:
Update Assembly runner to read only keys from current configured distribution country.


fredrb:
Shouldn't we run the distribution for keys that have the distribution country in the visited countries? From my understanding, country of origin holds the information of what country/app generated this keys, and not necessary to what country it is valid. 

For example, if we have the following keys: 

|Id|Origin|Visited Countries|
|-|-|-|
|1|DE|DE, FR|
|2|FR|FR, DK|
|3|DK|DK, FR|

Once we run the distribution for `FR` we want to select all the keys, not only the ones that were received from France.

sorin-stefan-iovita-sap:
Yes, that's correct. I missed that...

PR #719, opened 2020-08-20
Federation Gateway Upload Service

fredrb:
This is the initial version of Upload Service. In this PR we are introducing the following:
- Upload service module with a single dummy upload runner
- Flyway migrations to create the intermediate upload table, user and roles
- Initial version of key batch conversion of keys
- Upload Service Dockerfile and update docker-compose to run the upload job


common/persistence/src/main/resources/db/migration/V7__createFederationUploadKeyTable.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE federation_upload_key (
+    key_data bytea PRIMARY KEY,
+    rolling_period integer NOT NULL,
+    rolling_start_interval_number integer NOT NULL,
+    submission_timestamp bigint NOT NULL,
+    transmission_risk_level integer NOT NULL,
+    consent_to_federation boolean NOT NULL DEFAULT FALSE,
+    origin_country varchar (2),
```
michael-burwig, line 8: Are we expecting any other value than "DE" here at any point in time?

common/persistence/src/main/resources/db/migration/V7__createFederationUploadKeyTable.sql
```diff
@@ -0,0 +1,12 @@
...
+    visited_countries varchar (2) [],
+    report_type varchar(30),
```
michael-burwig, line 10: Same as above: Are we expecting a value set of size greater than 1 at any point in time?

common/persistence/src/main/resources/db/migration/V7__createFederationUploadKeyTable.sql
EugenM-SAP: You are right, some fields could be defaulted, including consent_to_federation, which is always true for uploadable keys. It was easier to clone the entire table for the db procedure which replicates the data on submission.  
Also, to point out here is that the implementation of the upload process uses the DiagnosisKey class to map an object to this particular table. This was mainly for ease of development, and because the two data structures (internal diagnosis key and uploadable diagnosis key) do not differ conceptually for this first release. They could diverge in the future though and we  could look into reducing the number of fields to the absolute required.

common/persistence/src/main/resources/db/migration/V7__createFederationUploadKeyTable.sql
EugenM-SAP: Are you referring to the visited Countries set? The answer would be yes

michael-burwig: I am referring to `report_type`. I assume the answer is the same as for `origin_country`.

EugenM-SAP: That's correct

services/federation-upload/src/test/java/app/coronawarn/server/services/federation/upload/integration/DiagnosisKeyReplicationIT.java
```diff
@@ -0,0 +1,127 @@
...
+import app.coronawarn.server.common.protocols.external.exposurenotification.ReportType;
+import app.coronawarn.server.services.federation.upload.Application;
+
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = {Application.class}, initializers = ConfigFileApplicationContextInitializer.class)
+@ActiveProfiles({"integration-test"})
+@Tag("s3-integration")
+class DiagnosisKeyReplicationIT {
+
+    private static final String TEST_ORIGIN_COUNTRY = "DE";
```
mibrasap, line 52: Line indentation seems to be different to the other files (4 and 3 instead of 2 spaces?)

mibrasap (changes requested):

common/persistence/src/main/java/app/coronawarn/server/common/persistence/repository/FederationUploadKeyRepository.java
```diff
@@ -0,0 +1,34 @@
...
+package app.coronawarn.server.common.persistence.repository;
+
+import app.coronawarn.server.common.persistence.domain.DiagnosisKey;
+import java.util.Collection;
+import org.springframework.data.jdbc.repository.query.Query;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface FederationUploadKeyRepository
+    extends org.springframework.data.repository.Repository<DiagnosisKey, Long> {
```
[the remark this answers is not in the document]
EugenM-SAP: The idea is that we use the DiagnosisKey class here just to be able to deserialize uploadable keys. The actual table is referenced in the native query. 
 If we were to switch to a CrudRepository, it would allow manipulation of the actual 'daignosis_key' table via generated methods .That's because spring repos use the entity metadata to find the concrete table.

services/federation-upload/src/test/java/app/coronawarn/server/services/federation/upload/integration/DiagnosisKeyReplicationIT.java
fredrb: I wonder if checkstyle validations are not working anymore also. Because this should've been caught by it.

common/persistence/src/main/resources/db/specific/postgresql/V9__createTriggersForUploadKeys.sql
```diff
@@ -0,0 +1,35 @@
...
+ 
+CREATE TRIGGER remove_expired_uploadable_keys_trigger
+    AFTER DELETE ON diagnosis_key
+    FOR EACH ROW EXECUTE PROCEDURE remove_expired_uploadable_keys();
+
+    
+-- Add the necessary permissions for the diagnosis key replication triggers
+
+GRANT SELECT, DELETE ON TABLE federation_upload_key TO "cwa_distribution";
+GRANT INSERT ON TABLE federation_upload_key TO "cwa_submission";
```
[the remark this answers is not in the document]
EugenM-SAP: There is a dedicated role for this table which is used to read (upload), then delete the records.
The other roles also need permissions, because:
 - cwa_submission will replicate keys to the upload table on each submission (the submission service does this)
 - cwa_distribution will delete keys when retention policies are applied. This will make sure that we never run into the scenario where keys which are too old (less than 14 days) are uploaded to the gateway.

common/persistence/src/main/java/app/coronawarn/server/common/persistence/repository/FederationUploadKeyRepository.java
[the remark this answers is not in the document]
mibrasap: Ok, fine for me

mibrasap (approved):

PR #725, opened 2020-08-24
Feature/federation downloader

mibrasap:
## Checklist


common/protocols/src/main/proto/app/coronawarn/server/common/protocols/external/exposurenotification/diagnosis_key_batch.proto
```diff
@@ -21,6 +21,7 @@ message DiagnosisKey {
   // E.g. 2 means TEK is 2 days after onset of symptoms.
   uint32 days_since_onset_of_symptoms = 8;
 }
+
```
fredrb, line 24: This is harmless to the protobuf spec, but useless in this PR. So could we remove to avoid future conflicts?

services/download/src/main/java/app/coronawarn/server/services/download/download/DiagnosisKeyBatchDownloader.java
```diff
@@ -0,0 +1,72 @@
...
+    try (Response response = federationGatewayClient.getDiagnosisKeys(
+        "application/json; version=1.0",
+        "abc",
+        "C=DE",
+        "2020-08-18")) {
+      DiagnosisKeyBatch.parseFrom(response.body().asInputStream());
+    } catch (IOException e) {
+      // TODO
+    }
+    return null;
```
fredrb, line 65: Are expecting to ever returning null in this method? Or is this just a placehold? If it can be null, I would prefer this method returns `Optional<DiagnosisKeyBatchContainer>` instead of `DiagnosisKeyBatchContainer` or `null`.

services/download/src/main/java/app/coronawarn/server/services/download/runner/Download.java
```diff
@@ -68,51 +71,54 @@
...
-        }
-      }
-
-    } catch (Exception e) {
-      logger.error("Download of diagnosis key batch failed.", e);
-      Application.killApplication(applicationContext);
+    // run();
+  }
+
+  private void run() {
```
fredrb, line 77: Why the private `run`? Also, the call to this method is commented on the overridden `run`. Is that on purpose?

services/download/src/test/java/app/coronawarn/server/services/download/FederationBatchRepositoryTest.java
```diff
@@ -0,0 +1,97 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.autoconfigure.web.reactive.AutoConfigureWebTestClient;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+@SpringBootTest
+@ExtendWith({SpringExtension.class})
+@AutoConfigureWebTestClient
+public class FederationBatchRepositoryTest {
```
fredrb, line 27: Shouldn't this test be in the persistence module? Since the class `FederationBatchRepository` is there.

services/download/src/test/resources/application.yaml
```diff
@@ -0,0 +1,29 @@
+---
+spring:
+  flyway:
+    enabled: true
+    locations: classpath:/db/migration
+  datasource:
+    url:  jdbc:tc:postgresql:11.8:///databasename?TC_TMPFS=/testtmpfs:rw
+#    if
+#    driver-class-name: org.testcontainers.jdbc.ContainerDatabaseDriver
```
fredrb, line 9: I know this comments are copied from other application.yaml files, but can we remove them from here? I guess they were there in the beginning while we were trying to use TestContainers.

common/federation/src/main/java/app/coronawarn/server/common/federation/client/FederationGatewayClient.java
```diff
@@ -20,28 +20,36 @@
...
-  String getDiagnosisKeys(@RequestHeader("Accept") String accept,
-      @RequestHeader("X-SSL-Client-SHA256") String shaClient, @RequestHeader("X-SSL-Client-DN") String dnClient,
-      @PathVariable("date") String date);
+  String getDiagnosisKeysString(@RequestHeader("Accept") String accept,
+                          @RequestHeader("X-SSL-Client-SHA256") String shaClient,
+                          @RequestHeader("X-SSL-Client-DN") String dnClient,
+                          @PathVariable("date") String date);
+
+  @GetMapping(value = "/diagnosiskeys/download/{date}")
+  Response getDiagnosisKeys(@RequestHeader("Accept") String accept,
```
fredrb, line 51: According to [this example](https://github.com/OpenFeign/feign#encoders) in FeignClient you can provide static headers to the Request. This way you don't have to always pass them as parameters. It would make the method cleaner:
```java
@GetMapping(value = "/diagnosiskeys/download/{date}")
@Headers({"key: value", "..."})
Response getDiagnosisKeys(@PathVariable("date") String date);
```

fredrb (changes requested):

fredrb:
One more thing, this PR adds a `FederationFeignHttpClientProvider` and our PR for the common client module (#727) adds a `FederationFeignHttpClientProvider`, which are exactly the same. Sonar is accusing duplication on that class also. This probably came from the earlier merge you did with the ssl branch. 
A suggestion would be to merge this branch with `feature/federation` and adjust to have only one of the classes.

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/FederationBatch.java
```diff
@@ -29,17 +29,23 @@
...
-  private Date date;
+  private LocalDate date;
 
   private FederationBatchStatus status;
 
   /**
    * Creates a FederationBatch.
    */
-  public FederationBatch(String batchTag, Date date, FederationBatchStatus status) {
+  public FederationBatch() {
```
EugenM-SAP, line 39: Because this is a core domain class, it would probably be good to add a description of what this entity represents.

EugenM-SAP: Is the default constructor needed? Could we try to preserve imutability of this class and use private final fields for the entity properties combined with a constructor which initalizez all fields? If yes, we could also remove the setters.

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/FederationBatchStatus.java
```diff
@@ -2,5 +2,6 @@
 
 public enum FederationBatchStatus {
```
EugenM-SAP, line 3: Just a suggestion to add a description of each status code and what it represents.

services/download/src/main/java/app/coronawarn/server/services/download/download/DiagnosisKeyBatchContainer.java
```diff
@@ -0,0 +1,39 @@
+package app.coronawarn.server.services.download.download;
+
+import app.coronawarn.server.common.protocols.external.exposurenotification.DiagnosisKeyBatch;
+
+public class DiagnosisKeyBatchContainer {
+
+  private final DiagnosisKeyBatch diagnosisKeyBatch;
+  private String batchTag;
```
EugenM-SAP, line 8: could the other fields be final as well, and remove the getters?

services/download/src/main/java/app/coronawarn/server/services/download/download/DiagnosisKeyBatchContainer.java
```diff
@@ -0,0 +1,39 @@
+package app.coronawarn.server.services.download.download;
+
+import app.coronawarn.server.common.protocols.external.exposurenotification.DiagnosisKeyBatch;
+
+public class DiagnosisKeyBatchContainer {
```
EugenM-SAP, line 5: Suggest to add a short description of what this data structure is intended for

services/download/src/main/java/app/coronawarn/server/services/download/download/DiagnosisKeyBatchDownloader.java
```diff
@@ -0,0 +1,110 @@
...
+import java.time.format.DateTimeFormatter;
+import java.util.Collection;
+import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Service;
+
+
+/**
+ * The BatchDownloader downloads the batches containing the keys.
```
EugenM-SAP, line 38: 'This class faciliates the downloading of diagnosis keys batches from the Federation Gateway Service"

services/download/src/main/java/app/coronawarn/server/services/download/runner/Download.java
```diff
@@ -0,0 +1,147 @@
...
+    processErrorFederationBatches();
+    
+    processFederationBatches();
+  }
+
+  private void saveFirstBatchTagForDate(LocalDate date) {
+    Optional<DiagnosisKeyBatchContainer> diagnosisKeyBatchContainerOptional =
+        diagnosisKeyBatchDownloader.downloadBatch(date);
+
+    if (diagnosisKeyBatchContainerOptional.isEmpty()) {
```
EugenM-SAP, line 83: You could use Optional.ifPresent( ...save batch ...)

services/download/src/main/java/app/coronawarn/server/services/download/runner/Download.java
```diff
@@ -0,0 +1,147 @@
...
+           DiagnosisKeyService diagnosisKeyService, DiagnosisKeyBatchDownloader diagnosisKeyBatchDownloader) {
+    this.federationBatchService = federationBatchService;
+    this.diagnosisKeyService = diagnosisKeyService;
+    this.diagnosisKeyBatchDownloader = diagnosisKeyBatchDownloader;
+  }
+
+  @Override
+  public void run(ApplicationArguments args) {
+    // TODO UTC handling?
+    LocalDate yesterday = LocalDate.now(); //LocalDate.now().minus(Period.ofDays(1));
```
EugenM-SAP, line 71: Should the yesterday computation be uncommented?

services/download/src/main/java/app/coronawarn/server/services/download/runner/Download.java
```diff
@@ -0,0 +1,147 @@
...
+      try {
+        Optional<DiagnosisKeyBatchContainer> diagnosisKeyBatchContainerOptional =
+            diagnosisKeyBatchDownloader.downloadBatch(federationBatch.getDate(), federationBatch.getBatchTag());
+
+        if (diagnosisKeyBatchContainerOptional.isEmpty()) {
+          federationBatchService.markFederationBatchWithStatus(federationBatch, FederationBatchStatus.ERROR_WONT_RETRY);
+        }
+
+        // store nextBatchTag
+        DiagnosisKeyBatchContainer diagnosisKeyBatchContainer = diagnosisKeyBatchContainerOptional.get();
```
EugenM-SAP, line 108: if the optional is empty the get will throw a runtime exception Maybe we can avoid that by using Optional.ifPresentOrElse

services/download/src/main/java/app/coronawarn/server/services/download/runner/Download.java
```diff
@@ -0,0 +1,147 @@
...
+
+    List<FederationBatch> federationBatchesWithErrors =
+        federationBatchService.findByStatus(FederationBatchStatus.ERROR);
+    federationBatchesWithErrors.forEach(federationBatch -> {
+      try {
+        Optional<DiagnosisKeyBatchContainer> diagnosisKeyBatchContainerOptional =
+            diagnosisKeyBatchDownloader.downloadBatch(federationBatch.getDate(), federationBatch.getBatchTag());
+
+        if (diagnosisKeyBatchContainerOptional.isEmpty()) {
+          federationBatchService.markFederationBatchWithStatus(federationBatch, FederationBatchStatus.ERROR_WONT_RETRY);
```
EugenM-SAP, line 104: The strategy is to retry downloading error batches only once?

services/download/src/main/java/app/coronawarn/server/services/download/runner/Download.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  private void processFederationBatches() {
+    // fetch one unprocessed (and no error state) federation batch
+    // store nextbatchTag, store diagnosis keys, update status to processed
+    // in case of error: set status to error
+    // loop
+
+    /*
+    FederationBatch batchToProcess = federationBatchService.getNextFederationBatchToProcess();
```
EugenM-SAP, line 139: Should this be uncommented?

services/download/src/main/java/app/coronawarn/server/services/download/runner/Download.java
mibrasap: Yes, that's the current approach

common/persistence/src/main/java/app/coronawarn/server/common/persistence/service/FederationBatchInfoService.java
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Sets the status of the provided federation batch.
+   */
+  public void updateStatus(FederationBatchInfo federationBatchInfo, FederationBatchStatus status) {
+    federationBatchInfoRepository
+        .saveDoUpdateStatusOnConflict(federationBatchInfo.getBatchTag(), federationBatchInfo.getDate(),
+            status.name());
+
+    logger.info("Marked batch with status.");
```
fredrb, line 63: Maybe add some more information to this log line? Perhaps what batch (tag) and status it was updated to?

common/persistence/src/main/resources/db/migration/V3__createTablesForDownload.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE federation_batch_info (
+    batch_tag   varchar(20) PRIMARY KEY,
```
fredrb, line 2: [EFGS uses varchar(64) for their batch fields](https://github.com/eu-federation-gateway-service/efgs-federation-gateway/blob/master/src/main/resources/db/changelog/v000-create-diagnosis-key-batch-entity-table.yml#L21-L26). Would be good to match that, just in case they start sending longer tags with hashed values or something similar.

common/persistence/src/main/resources/db/specific/postgresql/V4__createPermissionsForCallback.sql
```diff
@@ -1 +1 @@
-GRANT SELECT, INSERT ON TABLE federation_batch TO "cwa_federation_callback";
+GRANT SELECT, INSERT ON TABLE federation_batch_info TO "cwa_federation_callback";
```
fredrb, line 1: This migration was already applied in dev. Changing it means that we would need to reset the DB in dev manually. I guess it's not a big deal since it's only dev, but we should keep this in mind when merging this PR.

common/persistence/src/main/resources/db/migration/V3__createTablesForDownload.sql
michael-burwig: The specification is stating an even smaller value. (not sure how reliable the specification is though)

services/download/src/main/java/app/coronawarn/server/services/download/FederationBatchProcessor.java
```diff
@@ -0,0 +1,146 @@
...
+
+    while (!unprocessedBatches.isEmpty()) {
+      FederationBatchInfo currentBatch = unprocessedBatches.remove();
+      processBatchAndReturnNextBatchId(currentBatch, ERROR)
+          .ifPresent(nextBatchTag ->
+              unprocessedBatches.add(new FederationBatchInfo(nextBatchTag, currentBatch.getDate())));
+    }
+  }
+
+  private Optional<String> processBatchAndReturnNextBatchId(
```
fredrb, line 121: I find a bit confusing handling the update exception inside this method. Specially when one of the callers (`retryProcessingBatch`) also catches the error and does the error status update. Maybe for more clarity would be good to handle the exception only on the callers (`retryProcessingBatch` and `processUnprocessedFederationBatches`) and simply throw the exception here. 

If we need different log messages (or even different error handling per type of error) we should then use different Exception classes. So other than having a `try` block with a generic `Exception` catch, we could do something like:

```java
try {
  processBatchAndReturnNextBatchId(federationBatchInfo)
    // do something with the result
} catch (EFGSRequestException e)
  // ...
} catch (DBException e) {
  // ... 
}
```

This is more a refactoring suggestion rather than an issue. Based on what I understood from this part of the code I don't see any errors that could arise with the current implementation.

common/federation/src/main/java/app/coronawarn/server/common/federation/client/FederationGatewayClient.java
```diff
@@ -20,32 +20,34 @@
...
  */
 @FeignClient(name = "federation-server", url = "${federation-gateway.base-url}")
 public interface FederationGatewayClient {
 
-  @GetMapping(value = "/diagnosiskeys/download/{date}")
-  String getDiagnosisKeys(@RequestHeader("Accept") String accept,
-      @RequestHeader("X-SSL-Client-SHA256") String shaClient, @RequestHeader("X-SSL-Client-DN") String dnClient,
+  @GetMapping(value = "/diagnosiskeys/download/{date}",
+      headers = {"Accept=application/protobuf; version=1.0", "X-SSL-Client-SHA256=abcd", "X-SSL-Client-DN=C=PL"})
+  BatchDownloadResponse getDiagnosisKeys(@PathVariable("date") String date);
```
fredrb, line 46: I tested this locally and I noticed that when EFGS returns a 404 in this request for a day without batches. However the exception is thrown in the console directly, which makes it seem like something went wrong. Maybe we should implement error handling for the HTTP requests and log it properly with INFO when EFGS returns such calls.

fredrb (commented):
Overall it looks pretty good. Please check the comments inline and see if they are reasonable to implement in this PR. Otherwise I don't see them as necessary changes now.

common/federation/src/test/java/app/coronawarn/server/common/federation/client/download/FederationGatewayHttpMessageConverterTest.java
```diff
@@ -0,0 +1,124 @@
...
+    return buildHttpInputMessage(EXP_DIAGNOSIS_KEY_BATCH.toByteArray(), batchTag, nextBatchTag);
+  }
+
+  private static HttpInputMessage buildHttpInputMessage(byte[] body, String batchTag, String nextBatchTag)
+      throws IOException {
+    HttpInputMessage message = mock(HttpInputMessage.class);
+    when(message.getBody()).thenReturn(new ByteArrayInputStream(body));
+
+    HttpHeaders headers = new HttpHeaders();
+    headers.add(CONTENT_TYPE, "application/protobuf; version=1.0");
```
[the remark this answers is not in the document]
michael-burwig: I modified the `getSupportedMediaTypes` test to expect exactly `application/protobuf`.

common/persistence/src/main/java/app/coronawarn/server/common/persistence/repository/FederationBatchInfoRepository.java
```diff
@@ -20,29 +20,37 @@
...
+      @Param("status") String status);
 
-  @Query("SELECT * FROM federation_batch WHERE status != 'PROCESSED' OR status is null ORDER BY date asc LIMIT 1")
-  FederationBatch findOldestUnprocessedFederationBatch();
+  @Modifying
+  @Query("INSERT INTO federation_batch_info "
+      + "(batch_tag, date, status) "
+      + "VALUES (:batchTag, :date, :status) "
+      + "ON CONFLICT (batch_tag) DO UPDATE SET status=:status")
+  void saveDoUpdateStatusOnConflict(
```
hilmarf, line 50: do we need both `saveDoUpdateStatusOnConflict` vs. `saveDoNothingOnConflict` ?
isn't `saveDoUpdateStatusOnConflict` sufficient?

mibrasap: Yes, we need to ensure "DoNothingOnConflict", when new BatchInfos are stored coming in as nextBatchTag (or future callback). In this case, there might come in duplicates of batchTags, which should be ignored. On the other hand, we also need a way to update a status, which is the reason for "UpdateStatusOnConflict"

fredrb (approved):

PR #727, opened 2020-08-26
Feature/federation gateway configure ssl

EugenM-SAP:
## Checklist

## Description

New Maven module intended for common functionalities required by services which interact with the Federation Gateway.

* Declarative Feign web client for the Federation Gateway API (currently only contains the download keys operation - to be extended for other API resources)
* Configuration beans required to create the Feign clients
* Integration test to be added later when the local version of the FGS is available


common/federation/src/test/resources/application.yaml
```diff
@@ -0,0 +1,11 @@
+---
+logging:
+  level:
+    org:
+      springframework: off
+    root: off
+services:
```
fredrb: I guess we can remove this old properties

common/federation/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
fredrb, line 1: Do we need this file? We don't seem to have logs let alone tests in this module.

services/submission/src/main/java/app/coronawarn/server/services/submission/config/SubmissionServiceConfig.java
```diff
@@ -260,7 +260,8 @@ public Long getMonitoringBatchSize() {
   }
 
   public void setMonitoringBatchSize(Long batchSize) {
-    this.monitoring.setBatchSize(batchSize);
+    this.monitoring
```
fredrb, line 263: Can we revert this change? Doesn't seem to be relevant for this PR/feature.

fredrb (changes requested):

fredrb (approved):

Issue #700, opened 2020-08-11
Retention policy is applied 1 hour earlier


EvgeniiSkrebtcov:
## Describe the bug

As @Tho-Mat pointed out in https://github.com/corona-warn-app/cwa-server/issues/699#issuecomment-671358818, retention policy applied 1 hour earlier than it should:

> I also noticed that the index for (examples) the hour-file
> 
```
> 2020-07-26-hour-06.zip created 26.07.2020 09:05 (German Time)
> was removed at                 09.08.2020 08:05 (German Time)
> 
> 2020-07-26-hour-08.zip created 26.07.2020 11:05 (German Time)
> was removed at                 09.08.2020 10:05 (German Time)
```
> shouldn't they be removed one hour later to have them 14 Days on the server?

## Expected behavior

Retention policy should apply after 14 days.

fredrb:
The relevant part of the code for retention policy can be found [in this method](https://github.com/corona-warn-app/cwa-server/blob/ee78bf15ece7f282836149b5f39422892f055096/services/distribution/src/main/java/app/coronawarn/server/services/distribution/objectstore/S3RetentionPolicy.java#L60). It deletes all entries with the prefix `version/v1/diagnosis-keys/country/DE/date/<current date>`. This means that all keys for a single day should be deleted together. 

@Tho-Mat could you please share how did you determine the creation and removal dates?

Tho-Mat:
@fredrb 
If you download the hour or date files with "curl -R" you get the creation time/date of the files.
```
curl -R https://svc90.main.px.t-online.de/version/v1/diagnosis-keys/country/DE/date                 --output date.txt
curl -R https://svc90.main.px.t-online.de/version/v1/diagnosis-keys/country/DE/date/2020-08-13/hour --output hour.txt
```

And you are right, the hour-zip-files themselves are delete at the end of the day. 
But the index file "hour.txt" with the active hours "[0,1,2,3,4]" that point to the hour-zip-files  is updated one hour to early.

fredrb:
@Tho-Mat: I'm not sure I fully understand your point. I downloaded all the keys for today to check the creation days, and this was the result.

**Hour files:**
```bash
Aug 13 12:46 .
Aug 13 03:05 0
Aug 13 04:05 1
Aug 13 05:05 2
Aug 13 06:05 3
Aug 13 07:05 4
Aug 13 08:05 5
Aug 13 09:05 6
Aug 13 10:05 7
[...]

PR #711, opened 2020-08-14
Fix/700 retention policy

fredrb:
This pull request fixes issue #700 (internal tracking 104).

The retention policy for keys runs one hour too early. We were keeping 13 days and 23 hours of keys in the system other than full 14 days. 

For an overview of why this change was made, check [this comment and thread](https://github.com/corona-warn-app/cwa-server/issues/700#issuecomment-674057872).

PR #728, opened 2020-08-26
Federation/federation merged master

KevponSAP:


common/persistence/src/main/java/app/coronawarn/server/common/persistence/service/DiagnosisKeyService.java
```diff
@@ -133,9 +133,9 @@ public void applyRetentionPolicy(int daysToRetain, String countryCode) {
         .ofInstant(Instant.now(), UTC)
         .minusDays(daysToRetain)
         .toEpochSecond(UTC) / SECONDS_PER_HOUR;
-    int numberOfDeletions = keyRepository.countOlderThan(threshold, countryCode);
+    int numberOfDeletions = keyRepository.countOlderThanOrEqual(threshold, countryCode);
```
fredrb: What is the reason we are falling back to `countOlderThanOrEqual`? We needed to change this method to `countOlderThan` in PR #711 to fix issue #700.

fredrb (approved):

PR #730, opened 2020-08-27
Retrieve uploadable keys using new service in persistence module

EugenM-SAP:
## Checklist

## Description

Changes to support the provisioning of pending upload keys to the DiagnosisKeyBatchAssembler when the upload job is running. There will be follow-up PRs whch deal with the filtered selection of uploadable keys, and enforcing of batch constraints.

### DIFF
--- a/services/upload/src/main/java/app/coronawarn/server/services/federation/upload/runner/Upload.java
+++ b/services/upload/src/main/java/app/coronawarn/server/services/federation/upload/runner/Upload.java
@@ -1,9 +1,16 @@
 package app.coronawarn.server.services.federation.upload.runner;
 
+import app.coronawarn.server.common.persistence.domain.DiagnosisKey;
+import app.coronawarn.server.common.persistence.service.FederationUploadKeyService;
+import app.coronawarn.server.common.protocols.external.exposurenotification.DiagnosisKeyBatch;
+import app.coronawarn.server.services.federation.upload.Application;
+import app.coronawarn.server.services.federation.upload.DiagnosisKeyBatchAssembler;
+import java.util.List;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.boot.ApplicationArguments;
 import org.springframework.boot.ApplicationRunner;
+import org.springframework.context.ApplicationContext;
 import org.springframework.core.annotation.Order;
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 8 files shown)


services/upload/src/test/java/app/coronawarn/server/services/federation/upload/utils/MockData.java
```diff
@@ -0,0 +1,61 @@
...
+    List<DiagnosisKey> fakeKeys = new ArrayList<DiagnosisKey>();
+    while (numberOfKeys >= 0) {
+      fakeKeys.add(generateRandomDiagnosisKey(consentToShare));
+      numberOfKeys--;
+    }
+    return fakeKeys;
+  }
+
+  public static DiagnosisKey generateRandomDiagnosisKey(boolean consentToShare) {
+    DiagnosisKey dummyKey = DiagnosisKey.builder().withKeyData(randomByteData())
```
fredrb, line 43: Return the result of this call?

services/upload/src/main/java/app/coronawarn/server/services/federation/upload/runner/Upload.java
```diff
@@ -1,21 +1,45 @@
...
+    this.batchAssembler = batchAssembler;
+  }
 
   @Override
   public void run(ApplicationArguments args) throws Exception {
     logger.info("Running Upload Job");
+    try {
+      List<DiagnosisKey> pendingUploadKeys = uploadKeyService.getPendingUploadKeys();
+      List<DiagnosisKeyBatch> batches = batchAssembler.assembleDiagnosisKeyBatch(pendingUploadKeys);
+    } catch (Exception e) {
```
fredrb, line 40: None of these methods seem to throw an exception. Are you wrapping around a try/catch as a placeholder for future exceptions?

fredrb (commented):
I suspect the codestyle checks are not being run. This is not related to this PR as I think it didn't run on the `feature/federation` PRs either. You can see that the some of the files you added in the persistence module don't have imports ordered alphabetically, some have spaces between them and license is missing.  

Apart from that I think the code looks good, just two comments from my side.

services/upload/src/test/java/app/coronawarn/server/services/federation/upload/utils/MockData.java
EugenM-SAP: > I suspect the codestyle checks are not being run. This is not related to this PR as I think it didn't run on the `feature/federation` PRs either. You can see that the some of the files you added in the persistence module don't have imports ordered alphabetically, some have spaces between them and license is missing.
> 
> Apart from that I think the code looks good, just two comments from my side.

Right, currently checkstyles is not run on the federation branch.

services/upload/src/main/java/app/coronawarn/server/services/federation/upload/runner/Upload.java
EugenM-SAP: Right, at this point the Upload logic is incomplete. Next steps is to provide the batches to the http upload logic. However in any case, any runtime exception that could be thrown in this runner should be caught, and the cron job stopped immediately. This fail-fast logic is consistent with the distribution runners.

fredrb (approved):

PR #731, opened 2020-08-28
Feature/upload gateway integration

fredrb:
This PR adds a few things:

- Initial version of the signing classes (I thought what we have now would be sufficient, but after some investigation the signing needs to be done differently). 
- Wires together batch creation, signing and the call to the federation client
- Generates test data if `testdata` profile is active. This was used to test the scenario before having the read from DB implemented. However I think we can switch to a Test Data Generation similar to the one in Distribution in the future. Where the first runner step (provided that `testdata` profile is active) will generate and store test data in the Upload table.
- Doesn't call Federation Client if `fake-client` is active. This will log `keys`, `batchTag` and `batchSignature` on the console. 
- Fake private key to be used in the test environment

Keep in mind that `batchTag` and `batchSignature` are not proper values yet. The logic to generate both of them needs to be implemented.


common/federation/src/main/java/app/coronawarn/server/common/federation/client/FederationGatewayClient.java
```diff
@@ -44,4 +46,15 @@
...
\ No newline at end of file
+
+  @PostMapping(value = "/diagnosiskeys/upload")
+  @Headers({"Accept: application/json; version=1.0",
+      "Content-Type: application/protobuf; version=1.0",
+      "X-SSL-Client-SHA256: abc",
+      "X-SSL-Client-DN: C=DE"
+  })
+  String postBatchUpload(
+      byte[] raw,
```
EugenM-SAP, line 57: A javadoc with description of parameters would be good

services/federation-upload/src/test/resources/application.yaml
```diff
@@ -0,0 +1,27 @@
+---
+logging:
+  level:
+    org:
+      springframework: off
+    root: off
+
+services:
+  upload:
+    privatekey: ${VAULT_FILESIGNING_SECRET:file:/Users/i850193/workspaces/cwa/cwa-server/docker-compose-test-secrets/private.pem}
```
EugenM-SAP, line 10: the default of the property can be removed or pointed to a classpath pem fie src/test/resources

services/upload/src/main/java/app/coronawarn/server/services/federation/upload/keys/DiagnosisKeyGenerator.java
```diff
@@ -0,0 +1,38 @@
...
+        .withReportType(ReportType.CONFIRMED_CLINICAL_DIAGNOSIS)
+        .withConsentToFederation(true)
+        .withCountryCode("DE")
+        .withVisitedCountries(List.of("DE"))
+        .build();
+  }
+
+  @Override
+  public List<DiagnosisKey> loadDiagnosisKeys() {
+    return IntStream.rangeClosed(0, 250)
```
EugenM-SAP, line 34: maybe the number of keys to generate could be a parameter provided by the client

services/upload/src/main/java/app/coronawarn/server/services/federation/upload/payload/UploadPayload.java
```diff
@@ -0,0 +1,44 @@
+package app.coronawarn.server.services.federation.upload.payload;
+
+import app.coronawarn.server.common.protocols.external.exposurenotification.DiagnosisKeyBatch;
+
+/**
+ * This class represents an Upload call to the Federation Gateway. The payload for EFGS must contain the following
+ * information:
+ * - The bytes of the protobuf ({@link DiagnosisKeyBatch} batch).
```
EugenM-SAP, line 8: <li> items in the javadoc for better formatting

services/upload/src/main/java/app/coronawarn/server/services/federation/upload/runner/Upload.java
```diff
@@ -13,9 +19,39 @@
...
+    this.federationUploadClient.postBatchUpload(payload);
+  }
 
   @Override
   public void run(ApplicationArguments args) throws Exception {
     logger.info("Running Upload Job");
+    List<DiagnosisKey> diagnosisKeys = this.diagnosisKeyLoader.loadDiagnosisKeys();
+    logger.info("Generating Upload Payload for {} keys", diagnosisKeys.size());
+    List<UploadPayload> requests = this.payloadFactory.makePayloadList(diagnosisKeys);
+    logger.info("Executing {} batch request", requests.size());
```
EugenM-SAP, line 54: requests

services/upload/src/test/java/app/coronawarn/server/services/federation/upload/payload/helper/FakePrivateKeyResource.java
```diff
@@ -0,0 +1,36 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import java.io.IOException;
+import org.apache.commons.io.IOUtils;
+import org.springframework.core.io.Resource;
+import org.springframework.core.io.ResourceLoader;
+
+public class FakePrivateKeyResource {
+
+  private static String TEST_PRIVATE_KEY = "-----BEGIN EC PRIVATE KEY-----\n"
```
EugenM-SAP, line 13: Just a suggestion, maybe this could be moved in test files under src/test/resources

services/federation-upload/src/test/resources/application.yaml
fredrb: This seems to be the old file its path is under `federation-upload`. I must have accidentally pushed it. The proper one `upload/src/test/resources/application.yaml` points to a classpath file on the resources folder.

EugenM-SAP (approved):

PR #747, opened 2020-09-08
Feature/federation dl retention policy

mibrasap:
## Checklist

## Description

Delete Federation Batch Infos that are older than a specified amount of days (by default 14)


services/download/src/main/java/app/coronawarn/server/services/download/DownloadServiceConfig.java
```diff
@@ -34,6 +34,9 @@
   @Min(0)
   @Max(14)
   private Integer efgsOffsetDays;
+  @Min(0)
+  @Max(28)
```
hilmarf, line 38: who defined the 28?

hilmarf (approved):

services/download/src/main/java/app/coronawarn/server/services/download/DownloadServiceConfig.java
mibrasap: That's the same values as we have for the distributions service's retention policy

PR #789, opened 2020-09-17
Catch empty response in efgs download request

mibrasap:
Catch empty response in efgs download request

### DIFF
--- a/services/download/src/main/java/app/coronawarn/server/services/download/FederationBatchProcessor.java
+++ b/services/download/src/main/java/app/coronawarn/server/services/download/FederationBatchProcessor.java
@@ -76,7 +76,8 @@ public class FederationBatchProcessor {
   public void saveFirstBatchInfoForDate(LocalDate date) {
     try {
       logger.info("Downloading first batch for date {}", date);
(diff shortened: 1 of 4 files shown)


services/download/src/main/java/app/coronawarn/server/services/download/FederationBatchProcessor.java
```diff
@@ -124,8 +127,7 @@ public void processUnprocessedFederationBatches() {
     String batchTag = batchInfo.getBatchTag();
     logger.info("Processing batch for date {} and batchTag {}", date, batchTag);
     try {
-      BatchDownloadResponse response = federationGatewayClient.getDiagnosisKeys(batchTag, date);
-
+      BatchDownloadResponse response = federationGatewayClient.getDiagnosisKeys(batchTag, date).orElseThrow();
```
EugenM-SAP, line 130: This throws a NoSuchElementException if no response from EFGS which will kill the download process right? Will there be an error / failure log because of this?

mibrasap: Yes, there is logging in the catch-block accordingly

hilmarf (approved):

EugenM-SAP (approved):

PR #784, opened 2020-09-16
Feature/process upload response

ioangut:
## Checklist

* [x] Test your changes as thoroughly as possible before you commit them. Preferably, automate your test by unit/integration tests.
* [x] Make sure `mvn install` runs for the whole project and, if you touched any code in the respective service, submission and distribution service can be run with `spring-boot:run`

## Description

* Handle different EFGS responses for Upload Job as per [jira item](https://jira.itc.sap.com/browse/EXPOSUREBACK-160).

PR #790, opened 2020-09-17
use random base64 value to identify unique Upload runs

EugenM-SAP:
## Checklist

## Description

Fix needed for dry run testing. Cherry picked from https://github.com/corona-warn-app/cwa-server/pull/784


hilmarf (approved):
why do we use random byte and not e.g. UUID?

sorin-stefan-iovita-sap (approved):

EugenM-SAP:
> why do we use random byte and not e.g. UUID?

Hilmar, i would assume that at this point the UUID would be too large for the batch tag column size. But that's a question for Frederico

EvgeniiSkrebtcov:
> why do we use random byte and not e.g. UUID?

We should consider using a sequential numbering prefixed with a pod ID. Will give some traceability batchtag -> pod.

PR #801, opened 2020-09-22
Feature/log submission keys

mibrasap:
## Description

Bring the latest changes regarding checking and logging of the submitted diagnosis keys to feature/federation

### DIFF
--- a/common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/DiagnosisKey.java
+++ b/common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/DiagnosisKey.java
@@ -246,7 +246,7 @@ public class DiagnosisKey {
         + ", transmissionRiskLevel=" + transmissionRiskLevel
         + ", submissionTimestamp=" + submissionTimestamp
         + ", consentToFederation=" + consentToFederation
-        + ", originCountry='" + originCountry + '\''
+        + ", originCountry=" + originCountry
         + ", visitedCountries=" + visitedCountries
         + ", reportType=" + reportType
         + ", daysSinceOnsetOfSymptoms=" + daysSinceOnsetOfSymptoms
--- a/common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/DiagnosisKey.java
+++ b/common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/DiagnosisKey.java
@@ -236,4 +236,20 @@ public class DiagnosisKey {
     result = 31 * result + Arrays.hashCode(keyData);
     return result;
   }
+
+  @Override
+  public String toString() {
+    return "DiagnosisKey{"
+        + "keyData=HIDDEN"
+        + ", rollingStartIntervalNumber=" + rollingStartIntervalNumber
+        + ", rollingPeriod=" + rollingPeriod
+        + ", transmissionRiskLevel=" + transmissionRiskLevel
+        + ", submissionTimestamp=" + submissionTimestamp
+        + ", consentToFederation=" + consentToFederation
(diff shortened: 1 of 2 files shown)


common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/DiagnosisKey.java
```diff
@@ -236,4 +236,20 @@ public int hashCode() {
...
+  @Override
+  public String toString() {
+    return "DiagnosisKey{"
+        + "keyData=HIDDEN"
+        + ", rollingStartIntervalNumber=" + rollingStartIntervalNumber
+        + ", rollingPeriod=" + rollingPeriod
+        + ", transmissionRiskLevel=" + transmissionRiskLevel
+        + ", submissionTimestamp=" + submissionTimestamp
+        + ", consentToFederation=" + consentToFederation
+        + ", originCountry='" + originCountry + '\''
```
hilmarf, line 249: if you need to add ' and use string+ anyhow, why don't use "'" then? reading escape sequences, is often a hard task

hilmarf (dismissed):

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/DiagnosisKey.java
mibrasap: Thanks for the hint, it's fixed

services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java
```diff
@@ -145,6 +159,27 @@
     return diagnosisKeys;
   }
 
+  private void checkDiagnosisKeysStructure(List<DiagnosisKey> diagnosisKeys) {
+    diagnosisKeys.sort(Comparator.comparing(DiagnosisKey::getRollingStartIntervalNumber));
+    String keysString = Arrays.toString(diagnosisKeys.toArray());
+    Predicate<DiagnosisKey> hasRiskLevel6 = diagnosisKey -> diagnosisKey.getTransmissionRiskLevel() == 6;
+
+    if (diagnosisKeys.stream().noneMatch(hasRiskLevel6)) {
```
EugenM-SAP, line 167: the method name says checkDiagnosisKeysStructure which is more general. But in the implementation only risk 6 keys are being checked. Shouldn't we have logging for all types risk levels which would be expected in the payloads but are missing?

hilmarf (approved):

hilmarf (approved):

services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java
mibrasap: The number of keys to be submitted depends on the time how long the app is installed on the phone. Hence, the number of keys differs, but the key with transmission risk level 6 is the one from the day before submission and should always be present.

EugenM-SAP: I understand. It's worth adding a javadoc with your explanation

EugenM-SAP (approved):

EugenM-SAP (dismissed):

services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java
mibrasap: done ;)

hilmarf (approved):

EugenM-SAP (approved):

PR #804, opened 2020-09-22
On demand normalization of field values before persisting diagnosis keys

EugenM-SAP:
## Checklist

## Description

PR consists of the following: 

*  Added DiagnosisKeyNormalizer and other interfaces for applying normalization of field values during the construction of a Diagnosis Key. This will be an optional step which can be configured in the DiagnosisKeyBuilder via the method withFieldNormalizer()
* Added application.yml configuration for managing the mapping between DSOS (days since onset of symptoms) and TRL (transmission risk levels) used in deriving DSOS when it is missing from submission payloads. The values for this mapping are provided as dummy values just for testing purpose.  They are subject to change in the following weeks as the calculations are refined.
* Added a concrete DiagnosisKeyNormalizer implementation in the Submission module, which currently is able to derive DSOS from TRL
* Unit test for verifying that DSOS values are correctly derived


services/submission/src/main/resources/application.yaml
```diff
@@ -36,6 +36,16 @@ services:
...
       # The default 'origin country' when not provided in submission payload
       default-origin-country: ${ORIGIN_COUNTRY:DE}
+      tek-field-derivations:
+        # Map containing the derived values of DSOS (days since onset of symptoms) from TRL (transmission risk level)
+        # The values below are subject to change
+        dsos-from-trl:
+          1: -4
+          3: -3
+          5: -2
+          6: -1
```
fredrb, line 46: Isn't 6 the youngest key? So in this case it should be `6: 0` and `8: -1`?

services/submission/src/main/java/app/coronawarn/server/services/submission/normalization/SubmissionKeyNormalizer.java
```diff
@@ -0,0 +1,60 @@
...
+
+  @Override
+  public NormalizableFields normalize(NormalizableFields fieldsAndValues) {
+    int trlValue = fieldsAndValues.getTransmissionRiskLevel();
+    int dsos = fieldsAndValues.getDaysSinceOnsetOfSymptoms();
+
+    throwIfAllRequiredFieldsMissing(trlValue, dsos);
+
+    if (isMissing(dsos)) {
+      dsos = dsosFromTrlMap.getOrDefault(trlValue, 0);
```
fredrb, line 44: Wouldn't this set `dsos` to negative values? Since the mapping in `application.yaml` is from 0 to -4. I wonder if this is fine, since the guard check above `isMissing` validates that values are greater than 0.

services/submission/src/main/java/app/coronawarn/server/services/submission/normalization/SubmissionKeyNormalizer.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    return NormalizableFields.of(trlValue, dsos);
+  }
+
+  private void throwIfAllRequiredFieldsMissing(int trlValue, int dsos) {
+    if (isMissing(trlValue) && isMissing(dsos)) {
+      throw new IllegalArgumentException("Normalization of key values failed. A key was provided with"
+          + " both 'transmission risk level' and 'days since onset of symptoms' fields missing");
```
fredrb, line 53: Bit of a nitpick, but since this is an exception message, shouldn't we put the actual property names, rather than a full name? 
i.e. `transmissionRiskLevel` rather than 'transmission risk level'

fredrb (changes requested):
Codewise looks pretty good to me. Just want to clarify some things that I don't understand on the concept of normalization:

What I understood of the current flow is:

1. Submission service registers a `DiagnosisKeyNormalizer` object to `DiagnosisKeyBuilder`
2. Submission receives a payload 
3. After validations, payload is mapped to our domain class `DiagnosisKey`.
4. During this process, the builder calls the `Normalizer` class supplied by Submission
5. The `SubmissionKeyNormalizer` will then set the DSOS (if missing) based on the TRL values. 

Is this enough? It seems a bit too simplistic to just set a default if value is missing. Don't we also have to normalize the values set by the Mobile devices also? Or is this going to be done on the Upload? Since the TRL and DSOS calculation is particular to each app (as far as I know). 

I'm also struggling a bit to understand the overlap between normalization and input validation, and why it was done like this for submission. I thought we would need to normalize the keys coming and going to EFGS (upload and download). Where having or not DSOS in our own app is more in our control.

Maybe I'm missing the bigger picture, but based on your comment that normalization is contextual to each service, I suppose there is more coming on this topic to other services, right?

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/DiagnosisKeyBuilder.java
```diff
@@ -144,16 +147,28 @@ public FinalBuilder withDaysSinceOnsetOfSymptoms(int daysSinceOnsetOfSymptoms) {
     return this;
   }
 
+  @Override
+  public FinalBuilder withFieldNormalization(DiagnosisKeyNormalizer fieldNormalizer) {
```
hilmarf, line 151: some helpful JavaDoc maybe?

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/DiagnosisKeyBuilder.java
```diff
@@ -170,4 +185,12 @@ private DiagnosisKey throwIfValidationFails(DiagnosisKey diagnosisKey) {
 
     return diagnosisKey;
   }
+
+  private NormalizableFields normalizeValues() {
```
hilmarf, line 189: some helpful JavaDoc maybe?

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/normalization/NormalizableFields.java
```diff
@@ -0,0 +1,50 @@
...
+import app.coronawarn.server.common.persistence.domain.DiagnosisKey;
+
+/**
+ * This data structure is just a container for all fields of a
+ * {@link DiagnosisKey} to which normalization can be applied.
+ */
+public final class NormalizableFields {
+
+  private final int transmissionRiskLevel;
+  private final int daysSinceOnsetOfSymptoms;
```
hilmarf, line 32: if we would use `Integer` objects instead, it's easier to also have mappings available for `0`, because we can check on `null` object

services/submission/src/main/java/app/coronawarn/server/services/submission/normalization/SubmissionKeyNormalizer.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+  }
+
+  private boolean isMissing(int fieldValue) {
+    return fieldValue <= 0;
```
hilmarf, line 58: valid range for `daysSinceOnsetOfSymptoms` is -14 .. 0 .. 14 as far as I know...

services/submission/src/main/java/app/coronawarn/server/services/submission/normalization/SubmissionKeyNormalizer.java
```diff
@@ -0,0 +1,60 @@
...
+  public SubmissionKeyNormalizer(SubmissionServiceConfig config) {
+    dsosFromTrlMap = config.getTekFieldDerivations().getDsosFromTrl();
+  }
+
+  @Override
+  public NormalizableFields normalize(NormalizableFields fieldsAndValues) {
+    int trlValue = fieldsAndValues.getTransmissionRiskLevel();
+    int dsos = fieldsAndValues.getDaysSinceOnsetOfSymptoms();
+
+    throwIfAllRequiredFieldsMissing(trlValue, dsos);
```
hilmarf, line 41: who is handling the `IllegalArgumentException` then? what's happening in this cases?

hilmarf (changes requested):

services/submission/src/main/java/app/coronawarn/server/services/submission/normalization/SubmissionKeyNormalizer.java
```diff
@@ -0,0 +1,60 @@
...
+
+    if (isMissing(dsos)) {
```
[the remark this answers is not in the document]
EugenM-SAP: Yes it could, but that will be dealt with in another PR that is related to validating the input. 
For example, in case both fields are missing the payload will be rejected. The situation is under discussion

services/submission/src/main/resources/application.yaml
EugenM-SAP: These are just dummy values for now, just for testing purpose. The actual mapping values will be provided later, and i think we will need some changes in the derivation logic (currently in SubmissionKeyNormalizer) which is simplistic at the moment and does not take into account the day of the TEK. There are still discussions ongoing about how the mapping should look like, so there will be follow up PRs which refine this.

services/submission/src/main/java/app/coronawarn/server/services/submission/normalization/SubmissionKeyNormalizer.java
EugenM-SAP: That's a good point. I think the check for missing field should strictly verify the 0 value.

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/DiagnosisKeyBuilder.java
EugenM-SAP: The javadoc was added to the inherited interface. See FinalBuilder

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/normalization/NormalizableFields.java
EugenM-SAP: I agree it would be good, but the problem is the TEK data structure does not have Integer as data types for its fields. This means that we wouldn't know if the mobile sent 0 or did not send an int field at all.
See the generated protobuf specification class TemporaryExposureKey

services/submission/src/main/java/app/coronawarn/server/services/submission/normalization/SubmissionKeyNormalizer.java
EugenM-SAP: The error is mapped to a handler in ApiExceptionHandler

services/submission/src/main/java/app/coronawarn/server/services/submission/normalization/SubmissionKeyNormalizer.java
EugenM-SAP: Yes, i will change the logic, as i just realized that we can actually know if a field was provided or not in the TEK

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/normalization/NormalizableFields.java
EugenM-SAP: Nevermind the previous answer. After checking again, we can know if a field was set in the protobuf or not. 
I will make some changes to the missing field checks which you can review again

services/submission/src/main/java/app/coronawarn/server/services/submission/normalization/SubmissionKeyNormalizer.java
EugenM-SAP: On checking the TEK protobuf class again , i found that we can verify whether an optional field was provided or not, so i will change the implementation to better test this situatio. Initially i thought these integer fields are initialized with 0 in case not provided.

fredrb: sounds good

services/submission/src/main/resources/application.yaml
EugenM-SAP: > Codewise looks pretty good to me. Just want to clarify some things that I don't understand on the concept of normalization:
> 
> What I understood of the current flow is:
> 
> 1. Submission service registers a `DiagnosisKeyNormalizer` object to `DiagnosisKeyBuilder`
> 2. Submission receives a payload
> 3. After validations, payload is mapped to our domain class `DiagnosisKey`.
> 4. During this process, the builder calls the `Normalizer` class supplied by Submission
> 5. The `SubmissionKeyNormalizer` will then set the DSOS (if missing) based on the TRL values.
> 
> Is this enough? It seems a bit too simplistic to just set a default if value is missing. Don't we also have to normalize the values set by the Mobile devices also? Or is this going to be done on the Upload? Since the TRL and DSOS calculation is particular to each app (as far as I know).
> 
> I'm also struggling a bit to understand the overlap between normalization and input validation, and why it was done like this for submission. I thought we would need to normalize the keys coming and going to EFGS (upload and download). Where having or not DSOS in our own app is more in our control.
> 
> Maybe I'm missing the bigger picture, but based on your comment that normalization is contextual to each service, I suppose there is more coming on this topic to other services, right?

Your understanding of the flow in this implementation is correct.  Currently this is just a skeleton implementation of the whole normalization topic, in order to have a common ground  for the Download and Submission services to apply their respective (contextual) logic:
* The Download service will register another implementation of the normalizer interface and will apply the logic which translates the field values submitted by other national backends to our own specification.  That's in a follow up PR.

* The Submission normalizer implementation just derives DSOS and TRL from one another, in case they are missing. No other transformations are done. I guess this is where a confusion might arise, because what the current submission normalizer object does is rather input sanitization in some sense. This derivation logic and EFGS transformation rules  have been put under the same "normalization" roof somehow and that is reflected in the implementation. I thought of changing the interface to something along the lines of 'field transformation' to be more generic and include both normalization and sanitization but it didn't stick.

Hope that clarifies some things.

fredrb (approved):
That's fine with me.

hilmarf (approved):

PR #807, opened 2020-09-23
Redesign upload testdata generation

fredrb:
## Checklist

## Description

Reworks a bit on our Test Data Generation for Upload. Since we migrated the policies from distribution (i.e. key expiry check) our current mechanism was generating key that would be filtered out from Upload. With that in mind, we changed the following things:

1. Generate keys for day before
2. Only generate keys from latest generated key (avoid always generating a lot of keys every upload run)
3. Change the number test keys property from `keys` to `keys-per-hour` to reflect how the generation is being done. Other than having a big set of keys (e.g. 10k) and try to fit them in a 24h window (by splitting in keys per hours), it generates a smaller set of keys (e.g. 350) for each free hour from yesterday.

### DIFF
--- a/services/upload/src/main/java/app/coronawarn/server/services/federation/upload/runner/TestDataGeneration.java
+++ b/services/upload/src/main/java/app/coronawarn/server/services/federation/upload/runner/TestDataGeneration.java
@@ -24,6 +24,7 @@ import app.coronawarn.server.common.persistence.domain.DiagnosisKey;
 import app.coronawarn.server.common.persistence.domain.FederationUploadKey;
 import app.coronawarn.server.common.persistence.service.FederationUploadKeyService;
 import app.coronawarn.server.common.protocols.external.exposurenotification.ReportType;
+import app.coronawarn.server.common.protocols.internal.RiskLevel;
 import app.coronawarn.server.services.federation.upload.config.UploadServiceConfig;
 import app.coronawarn.server.services.federation.upload.testdata.TestDataUploadRepository;
 import java.security.SecureRandom;
@@ -60,8 +61,7 @@ public class TestDataGeneration implements ApplicationRunner {
   public static final long TEN_MINUTES_INTERVAL_SECONDS = TimeUnit.MINUTES.toSeconds(10);
 
   public TestDataGeneration(UploadServiceConfig uploadServiceConfig,
-      TestDataUploadRepository keyRepository,
-      FederationUploadKeyService federationUploadKeyService) {
+      TestDataUploadRepository keyRepository) {
     this.uploadServiceConfig = uploadServiceConfig;
     this.keyRepository = keyRepository;
   }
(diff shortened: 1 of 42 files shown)


services/upload/src/main/java/app/coronawarn/server/services/federation/upload/runner/TestDataGeneration.java
```diff
@@ -67,8 +72,8 @@ public TestDataGeneration(UploadServiceConfig uploadServiceConfig,
     return keydata;
   }
 
-  private DiagnosisKey makeKeyFromTimestamp(long timestamp) {
-    return DiagnosisKey.builder().withKeyData(randomByteData())
+  private FederationUploadKey makeKeyFromTimestamp(long timestamp) {
+    return FederationUploadKey.from(DiagnosisKey.builder().withKeyData(randomByteData())
         .withRollingStartIntervalNumber(generateRollingStartIntervalNumber(timestamp))
         .withTransmissionRiskLevel(2)
```
hilmarf, line 78: fixed value of `2`?

services/upload/src/main/java/app/coronawarn/server/services/federation/upload/runner/TestDataGeneration.java
```diff
@@ -77,7 +82,7 @@ private DiagnosisKey makeKeyFromTimestamp(long timestamp) {
         .withSubmissionTimestamp(timestamp)
         .withVisitedCountries(List.of("FR", "DK"))
```
hilmarf, line 83: shouldn't the list of countries be configurable or use the list of available countries?

hilmarf (dismissed):

services/upload/src/main/java/app/coronawarn/server/services/federation/upload/runner/TestDataGeneration.java
fredrb: The configuration for supported countries was done in distribution and submission, but not in upload. So we don't have this information here.

hilmarf (approved):

services/upload/src/main/java/app/coronawarn/server/services/federation/upload/runner/TestDataGeneration.java
```diff
@@ -67,17 +72,22 @@ public TestDataGeneration(UploadServiceConfig uploadServiceConfig,
...
         .withReportType(ReportType.CONFIRMED_TEST)
-        .build();
+        .build());
+  }
+
+  private int generateTransmissionRiskLevel() {
+    return Math.toIntExact(
+        getRandomBetween(RiskLevel.RISK_LEVEL_LOWEST_VALUE, RiskLevel.RISK_LEVEL_HIGHEST_VALUE));
```
EugenM-SAP, line 90: Not sure whether it's an issue or not that TRL values like 2 or 4 might be produced. Considering that the german app does not submit such values. Probably for testing it's not a problem

EugenM-SAP (approved):

services/upload/src/main/java/app/coronawarn/server/services/federation/upload/runner/TestDataGeneration.java
fredrb: Agreed. Maybe we have to think this through in the future. But distribution does it like this also. DEV and INT are filled with keys that have the wrong TRL.

PR #812, opened 2020-09-24
Add origin country to visited countries list if not already part of it

KevponSAP:
## Checklist

## Description

Fixes the issue that if the origin country is not part of visited countries in the submission payload the keys will never be distributed in the origin bucket since distribution service is based on visited countries.


emmetsap (dismissed):
Tests passed locally

services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java
```diff
@@ -126,12 +128,13 @@
...
+      List<String> visitedCountries = extendVisitedCountriesWithOriginCountry(
+          submissionPayload.getVisitedCountriesList());
 
       DiagnosisKey diagnosisKey = DiagnosisKey.builder()
           .fromTemporaryExposureKey(protoBufferKey)
-          .withVisitedCountries(submissionPayload.getVisitedCountriesList())
+          .withVisitedCountries(visitedCountries)
           .withCountryCode(originCountry)
           .withConsentToFederation(submissionPayload.getConsentToFederation())
           .withFieldNormalization(new SubmissionKeyNormalizer(submissionServiceConfig))
```
mibrasap, line 140: Did you consider extracting this to the DiagnosisKeyBuilder? Maybe it makes sense to concentrate the builder logic there for later reuse

mibrasap (changes requested):

services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java
KevponSAP: Might be difficult since it's part of another module where we don't have the origin country parameter

mibrasap (dismissed):

sorin-stefan-iovita-sap (dismissed):

sorin-stefan-iovita-sap (approved):

PR #813, opened 2020-09-24
Added initial TRL from DSOS derivation map and implementation

EugenM-SAP:
## Checklist

## Description

PR continues the work on normalization topic. It adds a dummy mapping used to derive TRL values from DSOS (as it is known so far) together with unit tests. The mapping and derivation logic will be refined in a follow-up PR.


fredrb (approved):
Looks good to me. I just wonder if we could put the mapping somewhere else other than the application.yaml. Not where or if it even makes sense. I just think that the map looks odd in the yaml.

emmetsap (approved):

EugenM-SAP:
> Looks good to me. I just wonder if we could put the mapping somewhere else other than the application.yaml. Not where or if it even makes sense. I just think that the map looks odd in the yaml.

I agree the lines are too long. I think we can do this after the mapping logic is clarified and we can find a more suitable place for it.

PR #819, opened 2020-09-25
Feature/federation empty response

mibrasap:
## Checklist

## Description

Adapt to an empty response from the EFGS server. Special cases may lead to a response "200 ok" with an empty response body. In this case, FeignClient is not utilizing the MessageConverter and hence, we have not been able to read the nextBatchTag. Now the FeignClient returns a ResponseEntity to a new service "FederationGatewayDownloadService" which takes care for handling empty responses and the nextBatchTag header.


services/download/src/main/java/app/coronawarn/server/services/download/FederationGatewayDownloadService.java
```diff
@@ -0,0 +1,106 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Component;
+
+/**
+ * Responsible for downloading and storing batch information from the federation gateway.
+ */
+@Component
+public class FederationGatewayDownloadService {
```
EugenM-SAP, line 39: The only comment i have is that you should consider to move this class to the common federation module.  Other wise all clients of this common module would need to reproduce this wrapper logic over the feign client in order for download to work properly. Either that, or the other way around, move any abstractions that are "higher" than the actual Feign client to the Download service module (like BatchDownloadResponse class <- not referenced anymore in the common module)

EugenM-SAP (approved):

services/download/src/main/java/app/coronawarn/server/services/download/FederationGatewayDownloadService.java
fredrb: The first option I don't see much value. Downloading batches is only in the domain of the download service. So I don't see any other consumers using this. But you raise a good point on the validity of having download specific code (i.e. the FeignClient) in the commons module in the first place. The second point I think this would be a bigger refactor/discussion and outside the scope of this PR. Because it would mean falling back to the bare version of the federation commons module. Where the only implementation in commons/federation was the Feign Client Configuration and even the interface was responsibility of each service. I guess that's still a valid approach as we see that there is only a small part of the EFGS integration needs to be shared.

services/download/src/main/java/app/coronawarn/server/services/download/FederationGatewayDownloadService.java
```diff
@@ -0,0 +1,106 @@
...
+   */
+  public BatchDownloadResponse downloadBatch(String batchTag, LocalDate date) {
+    String dateString = date.format(ISO_LOCAL_DATE);
+    try {
+      logger.info("Downloading batch for date {} and batchTag {}.", batchTag, dateString);
+      ResponseEntity<DiagnosisKeyBatch> response = federationGatewayClient
+          .getDiagnosisKeys(batchTag, dateString);
+      return parseResponseEntity(response);
+    } catch (FeignException e) {
+      logger.error("Downloading batch for date {} and batchTag {} failed.", batchTag, dateString);
```
fredrb, line 88: Here a failed batch is logged with its batch tag and date. But also where this method is called, the exception handling is also logging the error in a similar way:

**FederationBatchProcessor [138]:**
```java
logger.error("Federation batch processing for date {} and batchTag {} failed. Status set to {}",
          date, batchTag, errorStatus.name(), e);
```

services/download/src/test/java/app/coronawarn/server/services/download/DownloadIntegrationTest.java
```diff
@@ -26,52 +26,54 @@
...
+ * This integration test is responsible for testing the runners for download and retention policy. The Spring profile
+ * "federation-download-integration" enables the test data generation in /db/testdata/V99__createTestDataForIntegrationTest.sql
+ * via the application-federation-download-integration.yaml.
+ * <p>
+ * The sql script for the test data contains * a batch info for an expired batch that should be deleted by the retention
+ * policy * and two batch info from the current date of status 'ERROR', which should be reprocessed. One of them will be
+ * successfully reprocessed and the other one will fail. The WireMockServer is configured accordingly.
+ * <p>
+ * The WireMockServer will additionally return a series of three batches, where the first batch of the corresponding
+ * date is batch1, that  can be processed successfully. Batch2 is returned by an explicit call to its batch tag and can
```
fredrb, line 63:
```suggestion
 * date is batch1, that can be processed successfully. Batch2 is returned by an explicit call to its batch tag and can
```

services/download/src/test/java/app/coronawarn/server/services/download/DownloadIntegrationTest.java
```diff
@@ -26,52 +26,54 @@
...
- * * "batch2_tag" has state "PROCESSED"
- * * "batch3_tag" has state "ERROR"
- * * no batch has state "UNPROCESSED"
- *
+ * This integration test is responsible for testing the runners for download and retention policy. The Spring profile
+ * "federation-download-integration" enables the test data generation in /db/testdata/V99__createTestDataForIntegrationTest.sql
+ * via the application-federation-download-integration.yaml.
+ * <p>
+ * The sql script for the test data contains * a batch info for an expired batch that should be deleted by the retention
+ * policy * and two batch info from the current date of status 'ERROR', which should be reprocessed. One of them will be
```
fredrb, line 59: fix line comment prefixes (`*`)

services/download/src/test/java/app/coronawarn/server/services/download/DownloadIntegrationTest.java
```diff
@@ -26,52 +26,54 @@
...
+ * successfully reprocessed and the other one will fail. The WireMockServer is configured accordingly.
+ * <p>
+ * The WireMockServer will additionally return a series of three batches, where the first batch of the corresponding
+ * date is batch1, that  can be processed successfully. Batch2 is returned by an explicit call to its batch tag and can
+ * be processed successfully as well. Its successor batch3 fails with a 404 Not Found.
+ * <p>
+ * Hence, after the execution of both runners, the federation_batch_info table should be the following: *
+ * "expired_batch_tag" is deleted * "retry_batch_tag_successful" has state "PROCESSED" * "retry_batch_tag_fail" has
+ * state "ERROR_WONT_RETRY" * "batch1_tag" has state "PROCESSED" * "batch2_tag" has state "PROCESSED" * "batch3_tag" has
```
fredrb, line 68: fix line comment prefixes (`*`)

services/download/src/test/java/app/coronawarn/server/services/download/DownloadIntegrationTest.java
```diff
@@ -174,13 +179,10 @@ void testDownloadRunSuccessfully() {
     Iterable<DiagnosisKey> diagnosisKeys = diagnosisKeyRepository.findAll();
     assertThat(diagnosisKeys)
         .hasSize(3)
-        .contains(createDiagnosisKey(BATCH1_DATA))
-        .contains(createDiagnosisKey(BATCH2_DATA))
-        .contains(createDiagnosisKey(RETRY_BATCH_SUCCESSFUL_DATA));
+        .contains(FederationBatchTestHelper.createDiagnosisKey(BATCH1_DATA,downloadServiceConfig))
```
fredrb, line 182:
```suggestion
        .contains(FederationBatchTestHelper.createDiagnosisKey(BATCH1_DATA, downloadServiceConfig))
```

services/download/src/test/java/app/coronawarn/server/services/download/DownloadIntegrationTest.java
```diff
@@ -174,13 +179,10 @@ void testDownloadRunSuccessfully() {
...
+        .contains(FederationBatchTestHelper.createDiagnosisKey(BATCH2_DATA,downloadServiceConfig))
```
fredrb, line 183:
```suggestion
        .contains(FederationBatchTestHelper.createDiagnosisKey(BATCH2_DATA, downloadServiceConfig))
```

fredrb (changes requested):

services/download/src/main/java/app/coronawarn/server/services/download/FederationGatewayDownloadService.java
fredrb: The first option I don't see much value. Downloading batches is only in the domain of the download service. So I don't see any other consumers using this. But you raise a good point on the validity of having download specific code (i.e. the FeignClient) in the commons module in the first place. The second point I think this would be a bigger refactor/discussion and outside the scope of this PR. Because it would mean falling back to the bare version of the federation commons module. Where the only implementation in commons/federation was the Feign Client Configuration and even the interface was responsibility of each service. I guess that's still a valid approach as we see that there is only a small part of the EFGS integration needs to be shared.

mibrasap: Eugen and me talked yesterday a bit about this topic. BatchDownloadResponse is already moved from the commons module to the downloader. Good points with the shared parts of the FeignClant. Let's keep the discussion for the refactoring for the future.

services/download/src/main/java/app/coronawarn/server/services/download/FederationGatewayDownloadService.java
mibrasap: The first one is related to the actual download, the second one is for processing it, according to the different layers of FederationGatewayService and FederationBatchProcessor

EugenM-SAP (approved):

fredrb (approved):

PR #828, opened 2020-09-30
Rework Upload TestDataGeneration

fredrb:
If the upload fails, pending keys are left with an empty batch tag in the DB. This means that after many failed upload runs, we can get a pretty high amount of fake keys in the DB. 

This PR changes the TestDataGeneration to generate fake keys only until a threshold is met (currently set to 8k).

### DIFF
--- a/services/upload/src/main/java/app/coronawarn/server/services/federation/upload/runner/TestDataGeneration.java
+++ b/services/upload/src/main/java/app/coronawarn/server/services/federation/upload/runner/TestDataGeneration.java
@@ -31,15 +31,12 @@ import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZoneOffset;
 import java.time.temporal.ChronoUnit;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
-import java.util.concurrent.ThreadLocalRandom;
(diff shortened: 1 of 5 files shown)


hilmarf (approved):

services/upload/src/main/java/app/coronawarn/server/services/federation/upload/runner/TestDataGeneration.java
```diff
@@ -119,43 +113,37 @@ private long secondsToHours(long timestampInSeconds) {
...
+      long hourEnd = secondsToHours(upperHour.toEpochSecond(ZoneOffset.UTC));
+
+      logger.info("Generating {} fake upload keys between times {} and {}",
+          numberOfKeysToGenerate,
           hourStart,
           hourEnd);
-      return LongStream.range(hourStart, hourEnd)
-          .mapToObj(i -> this.makeKeysFromTimestamp(i, keysToGeneratePerHour))
-          .flatMap(Collection::stream)
+      var random = ThreadLocalRandom.current();
```
sorin-stefan-iovita-sap, line 141: Use SecureRandom instead. 🛡️

sorin-stefan-iovita-sap (approved):

mibrasap (approved):

PR #842, opened 2020-10-02
Validate downloadserviceconfig

JonasZeitzem-sap:
## Checklist

* [X] Test your changes as thoroughly as possible before you commit them. Preferably, automate your test by unit/integration tests.
* [X] Make sure `mvn install` runs for the whole project and, if you touched any code in the respective service, submission and distribution service can be run with `spring-boot:run`

## Description
Adds validation for the DownloadServiceConfiguration, more specifically for the map containing the derived values of TRL (transmission risk level) from DSOS (days since onset of symptoms).
Also included in this PR is the introduction of a new "config" package and unit tests for the introduced validations.


services/download/src/main/java/app/coronawarn/server/services/download/config/DownloadServiceConfigValidator.java
```diff
@@ -0,0 +1,49 @@
...
+package app.coronawarn.server.services.download.config;
+
+import java.util.Map;
+import org.springframework.validation.Errors;
+import org.springframework.validation.Validator;
+
+/**
+ * Validate the values of the DownloadServiceConfig.
+ */
+public class DownloadServiceConfigValidator implements Validator {
```
EugenM-SAP, line 12: I think you need to register this validator bean statically before the config properties get injected in the DownloadServiceConfig. See how the SubmissionServiceValidator is registered in the Application.class of the submission module

hilmarf (dismissed):

services/download/src/main/java/app/coronawarn/server/services/download/config/DownloadServiceConfigValidator.java
```diff
@@ -0,0 +1,49 @@
...
+
+  private void checkTransmissionRiskLevelInAllowedRange(Integer transmissionRiskLevel, Errors errors) {
+    if (transmissionRiskLevel > 8 || transmissionRiskLevel < 1) {
+      errors.rejectValue("tekFieldDerivations",
+          "[" + transmissionRiskLevel + "]: transmissionRiskLevel value is not in the allowed range (1 to 8)");
+    }
+  }
+
+  private void checkDaysSinceOnsetOfSymptomsInAllowedRange(Integer daysSinceOnsetOfSymptoms, Errors errors) {
+    if (daysSinceOnsetOfSymptoms > 4000 || daysSinceOnsetOfSymptoms < -14) {
```
fredrb, line 43: Checking for valid DSOS values is also done on [key level here](https://github.com/corona-warn-app/cwa-server/blob/master/services/download/src/main/java/app/coronawarn/server/services/download/validation/ValidFederationKeyFilter.java#L27-L35).

I wonder if we should have this range `-14, 4000` as a constant somewhere, like we for [specific ENF constants](https://github.com/corona-warn-app/cwa-server/blob/master/services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/appconfig/validation/ParameterSpec.java).

EugenM-SAP: Yes, we will need to either externalize these rangs somewhere (both TRL and DSOS) or centralize them in a constants class under a common module.  This needs ot be done for submission as well, but we are waiting for some clarifications regarding the actual values of dsos.

fredrb: Fair enough. I don't think this stops this PR, just wanted to raise this.

fredrb (approved):

services/download/src/main/java/app/coronawarn/server/services/download/config/DownloadServiceConfigValidator.java
JonasZeitzem-sap: Yes, agree with you here. This should be improved. Maybe even take this a step further an centralize the mapping between the two values completely not only the allowed ranges? For the Trl to Dsos mapping we already have two places to maintain it, if something changes.

services/download/src/main/java/app/coronawarn/server/services/download/config/DownloadServiceConfigValidator.java
JonasZeitzem-sap: Correct. Without the static bean definition the validation is not executed. Did the change as requested.

fredrb (approved):

EugenM-SAP (approved):

ioangut (approved):

PR #843, opened 2020-10-02
Feature/federation download filtering

mibrasap:


services/download/src/main/java/app/coronawarn/server/services/download/FederationBatchProcessor.java
```diff
@@ -116,18 +118,26 @@ public void processUnprocessedFederationBatches() {
...
       BatchDownloadResponse response = federationGatewayDownloadService.downloadBatch(batchTag, date);
+      AtomicBoolean batchContainsInvalidKeys = new AtomicBoolean(false);
       response.getDiagnosisKeyBatch().ifPresent(diagnosisKeyBatch -> {
-        logger.info("Downloaded {} keys for date {} and batchTag {}", diagnosisKeyBatch.getKeysCount(), date, batchTag);
-        int insertedKeys = diagnosisKeyService.saveDiagnosisKeys(extractValidDiagnosisKeys(diagnosisKeyBatch));
-        logger.info("Successfully inserted {} keys for date {} and batchTag {}", insertedKeys, date, batchTag);
+        logger
+            .info("Downloaded {} keys for date {} and batchTag {}.", diagnosisKeyBatch.getKeysCount(), date, batchTag);
+        List<DiagnosisKey> validDiagnosisKeys = extractValidDiagnosisKeys(diagnosisKeyBatch);
+        int invalidKeys = diagnosisKeyBatch.getKeysCount() - validDiagnosisKeys.size();
```
EugenM-SAP, line 129: Worth extracting a method here : logInvalidKeys()

services/download/src/main/java/app/coronawarn/server/services/download/FederationBatchProcessor.java
```diff
@@ -138,9 +148,18 @@ public void processUnprocessedFederationBatches() {
     return diagnosisKeyBatch.getKeysList()
         .stream()
         .filter(validFederationKeyFilter::isValid)
-        .map(diagnosisKey -> DiagnosisKey.builder().fromFederationDiagnosisKey(diagnosisKey)
-            .withFieldNormalization(new FederationKeyNormalizer(config))
-            .build())
+        .map(diagnosisKey -> {
+          try {
```
EugenM-SAP, line 152: worth to extract a method for trying to create the DiagnosisKey from the federation key and hide the try catch.

Maybe map(
                 Optional.of(  transformToDiagnosisKey(federationKey) ) 
            ....)

kilbphilippSAP (approved):

services/download/src/main/java/app/coronawarn/server/services/download/FederationBatchProcessor.java
```diff
@@ -116,31 +118,50 @@ public void processUnprocessedFederationBatches() {
...
   }
+
+  private Optional<DiagnosisKey> convertFederationDiagnosisKeyToDiagnosisKey(
+      app.coronawarn.server.common.protocols.external.exposurenotification.DiagnosisKey diagnosisKey) {
+    try {
+      return Optional.of(DiagnosisKey.builder().fromFederationDiagnosisKey(diagnosisKey)
+          .withFieldNormalization(new FederationKeyNormalizer(config))
+          .build());
+    } catch (Exception ex) {
+      logger.info("Building key failed.");
```
fredrb, line 163:
```suggestion
      logger.error("Building key failed.", ex);
```

EugenM-SAP (approved):

services/download/src/main/java/app/coronawarn/server/services/download/validation/ValidFederationKeyFilter.java
```diff
@@ -1,37 +1,123 @@
...
+    this.maxRollingPeriod = validation.getMaxRollingPeriod();
+    this.minTrl = validation.getMinTrl();
+    this.maxTrl = validation.getMaxTrl();
+  }
 
   /**
    * Accepts or rejects a key based on the evaluation of the fields against permitted values.
    */
   public boolean isValid(
       app.coronawarn.server.common.protocols.external.exposurenotification.DiagnosisKey federationKey) {
```
fredrb, line 52:
```suggestion
      DiagnosisKey federationKey) {
```
No longer needed as the `exposurenotification.DiagnosisKey` is the only imported class with `DiagnosisKey` name 🙂

services/download/src/main/java/app/coronawarn/server/services/download/validation/ValidFederationKeyFilter.java
```diff
@@ -1,37 +1,123 @@
...
+      logger.info("Filter skipped Federation DiagnosisKey with invalid 'KeyData' length {}.",
+          federationKey.getKeyData().toByteArray().length);
+    }
+    return hasCorrectKeyLength;
+  }
+
+  private boolean hasValidStartIntervalNumber(DiagnosisKey federationKey) {
+    boolean hasValidStartIntervalNumber = federationKey.hasRollingStartIntervalNumber()
+        // rolling start interval number is midnight UTC
+        && federationKey.getRollingStartIntervalNumber() % maxRollingPeriod == 0;
```
fredrb, line 93: I suggest you extract this into another method and remove the comment for clarify. For example:
```java
 && rollingStartIntervalIsMidnight(federationKey);
```

fredrb (changes requested):
I don't understand why some of this validations are being done in Download. The validation if a DiagnosisKey is valid or not should already be covered on the DiagnosisKey builder, and the definition of what is a valid field mapped in the properties of our Persistence Class. Take `rollingPeriod` for example, [DiagnosisKey](https://github.com/corona-warn-app/cwa-server/blob/master/common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/DiagnosisKey.java#L49-L51) domain class already annotates what are the valid values. 

Therefore, I feel like the "validation loop" does very similar validations twice in a row. It first filters all the keys that are valid, then calls the builder to create DiagnosisKeys. The build will work for every key because the invalid ones were already filtered out.  

```java
    return diagnosisKeyBatch.getKeysList()
        .stream()
        .filter(validFederationKeyFilter::isValid) // <-- Validates all fields
        .map(this::convertFederationDiagnosisKeyToDiagnosisKey)  // <-- Validates all fields again
```

This works, but I feel like it's misleading. Because sometime we rely on a separate validator class (i.e. ValidFederationKeyFilter) and sometime on the validations defined in our Model class (i.e. domain/DiagnosisKey). Maybe some validations are needed specifically on Download, but I don't think all of them are. 

The second thing about this is that we are now keeping validation ranges in multiple places:

1. [DiagnosisKey](https://github.com/corona-warn-app/cwa-server/blob/master/common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/DiagnosisKey.java#L37-L38) class has a static public constant for Rolling period and hardcodes other values. This PR adds the same values in `application.yaml`
2. #842 adds ranges for checking DSOS hardcoded. This PR added the same values in `application.yaml`. I raised the same point in the PR. We are having awfully similar values defined in many places. 

Maybe it's fine to have these values as constants as they won't change often (or ever), and avoid cluttering `application.yaml` with a lot of constants. But whatever we choose, we should always read from the same place.

PR #849, opened 2020-10-05
convert report type from int

fredrb:
Report type could not be converted from number to Enum directly.

### DIFF
--- a/services/upload/src/main/resources/application.yaml
+++ b/services/upload/src/main/resources/application.yaml
@@ -33,7 +33,7 @@ services:
       enable-dsos: ${ENABLE_DSOS_TRANSMISSION_TO_EFGS:true}
       enable-report-type: ${ENABLE_REPORT_TYPE_TRANSMISSION_TO_EFGS:true}
       default-dsos: ${DEFAULT_DSOS:1}
-      default-report-type: ${DEFAULT_REPORT_TYPE:0}
+      default-report-type: ${DEFAULT_REPORT_TYPE:UNKNOWN}
     signature:
       # The ASN.1 OID for algorithm identifier.
       algorithm-oid: 1.2.840.10045.4.3.2
--- a/services/upload/src/test/resources/application.yaml
+++ b/services/upload/src/test/resources/application.yaml
@@ -25,7 +25,7 @@ services:
       enable-dsos: true
       enable-report-type: true
       default-dsos: 1
-      default-report-type: 0
+      default-report-type: UNKNOWN
     signature:
       # The ASN.1 OID for algorithm identifier.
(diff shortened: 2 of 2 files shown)


services/upload/src/main/java/app/coronawarn/server/services/federation/upload/config/UploadServiceConfig.java
```diff
@@ -100,8 +100,8 @@ public ReportType getDefaultReportType() {
       return defaultReportType;
     }
 
-    public void setDefaultReportType(ReportType defaultReportType) {
-      this.defaultReportType = defaultReportType;
```
hilmarf, line 104: this isn't required then?

hilmarf (approved):

services/upload/src/main/java/app/coronawarn/server/services/federation/upload/config/UploadServiceConfig.java
fredrb: It's still required. We just have to consider that `application.yaml` has a number and our class expects a `ReportType`

EvgeniiSkrebtcov (dismissed):

fredrb:
@EvgeniiSkrebtcov @hilmarf: I have changed the approach after talking with @EugenM-SAP. Spring can convert the enum name from a String, which makes the solution a lot cleaner.

services/upload/src/main/resources/application.yaml
```diff
@@ -33,7 +33,7 @@ services:
       enable-dsos: ${ENABLE_DSOS_TRANSMISSION_TO_EFGS:true}
       enable-report-type: ${ENABLE_REPORT_TYPE_TRANSMISSION_TO_EFGS:true}
       default-dsos: ${DEFAULT_DSOS:1}
-      default-report-type: ${DEFAULT_REPORT_TYPE:0}
+      default-report-type: ${DEFAULT_REPORT_TYPE:UNKNOWN}
```
hilmarf, line 36: when is this value used? only in case ENABLE_REPORT_TYPE_TRANSMISSION_TO_EFGS == false, right?

hilmarf (approved):

EugenM-SAP (approved):

ioangut (approved):

sorin-stefan-iovita-sap (approved):

services/upload/src/main/resources/application.yaml
fredrb: Correct. This snippet shows how it's used:
```java
public ReportType getReportTypeOrDefault(ReportType type) {
    return canSendReportType ? type : defaultReportType;
}
```
This method is always called when we are converting the keys to be uploaded on a batch.

PR #854, opened 2020-10-06
Enhancement/origin contry validation

ioangut:
## Checklist

## Description

This PR brings into the payload validation one more condition for the originCountry such that if there is an empty originCountry field in the payload, it would not validate this field based on the supported country, but it would allow to set the default value.

### DIFF
--- a/services/submission/src/main/java/app/coronawarn/server/services/submission/validation/ValidSubmissionPayload.java
+++ b/services/submission/src/main/java/app/coronawarn/server/services/submission/validation/ValidSubmissionPayload.java
@@ -124,7 +124,7 @@ public @interface ValidSubmissionPayload {
     private boolean checkOriginCountryIsValid(SubmissionPayload submissionPayload,
         ConstraintValidatorContext validatorContext) {
       String originCountry = submissionPayload.getOrigin();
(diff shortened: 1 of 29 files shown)


services/submission/src/main/java/app/coronawarn/server/services/submission/validation/ValidSubmissionPayload.java
```diff
@@ -124,7 +124,8 @@ private boolean checkStartIntervalNumberIsAtMidNight(List<TemporaryExposureKey>
     private boolean checkOriginCountryIsValid(SubmissionPayload submissionPayload,
         ConstraintValidatorContext validatorContext) {
       String originCountry = submissionPayload.getOrigin();
-      if (submissionPayload.hasOrigin() && !supportedCountries.contains(originCountry)) {
+      if (submissionPayload.hasOrigin() && !submissionPayload.getOrigin().isEmpty()
```
hilmarf, line 127: use `!originCountry.isEmpty()` as it was created in line 126 already

hilmarf (changes requested):

hilmarf (approved):

EvgeniiSkrebtcov (approved):

EugenM-SAP (approved):

PR #884, opened 2020-10-13
Document EUR package

sorin-stefan-iovita-sap:
## Description
This PR documents the implementation about the EUR distribution package.


docs/DISTRIBUTION.md
```diff
@@ -113,6 +113,10 @@ date and hour directory also contain an index file each, listing all the subdire
 concept, but writing this structure to the local disk is not supported. Therefore, this country file will be assembled as
 `version/v1/diagnosis-keys/country/index`. The `/index` part will then be removed upon S3 upload.
 
+In the current implementation the `<ISO-3166-country>` placeholder is expected to have only `DE` and `EUR` values. The `DE`
+folder is containing the distributed keys for the German country and the `EUR` folder is containing the distributed keys for the whole
+Europe countries.
+
```
KevponSAP, line 119:
```suggestion
The `origin_country` (ISO-3166 country code) parameter defines which country folder is being created. The origin country folder contains only the distributable keys for the origin country. Furthermore, based on the parameter`eu-package-name` an additional folder will be created which provides all keys from the supported countries in a single package. 
```

KevponSAP (changes requested):

docs/DISTRIBUTION.md
```diff
@@ -113,6 +113,11 @@ date and hour directory also contain an index file each, listing all the subdire
 concept, but writing this structure to the local disk is not supported. Therefore, this country file will be assembled as
 `version/v1/diagnosis-keys/country/index`. The `/index` part will then be removed upon S3 upload.
 
+The `origin_country` (ISO-3166 country code) parameter defines which country folder is being created.
```
EugenM-SAP, line 116: Suggested rewording:
"...parameter refers to the country in which the CWA clients and backend operate. It is used to create a folder on the CDN distribution paths, where so called "national key packages" are distributed. In essence, these packages  contain keys submitted by CWA clients as well as keys from visitors of the country, that have been downloaded from the Federation Gateway Service."

fredrb:
Maybe we should include that on the origin country we apply all the key sharing policies? While on the EUR keys are distributed immediately?

KevponSAP (approved):

fredrb (approved):

PR #886, opened 2020-10-14
Fix testdata generator

KevponSAP:
## Checklist

## Description

- Setting the consentToFederation parameter to true while generating diagnosis keys via the distribution testdata generator
- Set default report type to CONFIRMED_TEST

### DIFF
--- a/services/distribution/src/main/java/app/coronawarn/server/services/distribution/config/DistributionServiceConfig.java
+++ b/services/distribution/src/main/java/app/coronawarn/server/services/distribution/config/DistributionServiceConfig.java
@@ -257,6 +257,8 @@ public class DistributionServiceConfig {
     private Integer seed;
     private Integer exposuresPerHour;
 
+    private boolean consentToFederation;
+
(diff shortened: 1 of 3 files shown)


sorin-stefan-iovita-sap (approved):

EugenM-SAP (approved):
If the intent is to share the keys generated here with EFGS during testing, it's fine.

fredrb (changes requested):
I'm fine with the code. I would just suggest to change the environment variable name to something more specific. For example `TESTDATA_CONSENT_TO_FEDERATION` or even `DISTRIBUTION_TESTDATA_CONSENT_TO_FEDERATION`. To make it clear that this is used only for test data.

fredrb (approved):

EugenM-SAP (approved):
Maybe just a remark, that the env could be even more specific.
"SHARE_TESTDATA_WITH_EFGS" true/false
In the end that's what the goal is.

kilbphilippSAP (approved):

PR #893, opened 2020-10-16
Set reportType to ConfirmedTest for each downloaded key

mibrasap:
Apply the hotfix also to master


sorin-stefan-iovita-sap (approved):

ioangut (approved):

fredrb (changes requested):
The commits from this PR overlap with commits from the "merge to master PR" #894. Could we maybe close this PR and merge only the other one? Otherwise we will get conflicts in files we already applied this patch.

PR #898, opened 2020-10-20
Introduce new paramter that allows to enforce downloading batches for…

mibrasap:
## Checklist

* [ x ] Test your changes as thoroughly as possible before you commit them. Preferably, automate your test by unit/integration tests.
* [ x ] Make sure `mvn install` runs for the whole project and, if you touched any code in the respective service, submission and distribution service can be run with `spring-boot:run`

## Description

The Federation Batch Info stores information about which batches have already been processed to not download them again. The batches for the current day might change constantly when national backends upload keys, hence there is the need to download the batches for today again. Hence, the entries in federation batch info with the current day need to be removed. There is a parameter 'efgs-repeat-download-offset-days' with default 0 for that.

### DIFF
--- a/services/download/src/main/resources/application.yaml
+++ b/services/download/src/main/resources/application.yaml
@@ -14,8 +14,9 @@ services:
     # the keys for today. If it is set to 1, the keys from yesterday will be downloaded.
     # The value must be between 0 and 14.
     efgs-offset-days: ${EFGS_OFFSET_DAYS:1}
-    # The offset in days for which the keys from the federation gateway shall be downloaded from,
-    # ignoring if a batch of this day has already been downloaded.
+    # This is the offset in past days for which keys should be downloaded again, regardless if a
+    # previous download had occurred. I.e. Value '3' means re-download all batches/keys from 3 days ago.
+    # Note that the re-download happens only if this value matches the efgs-offset-days.
     efgs-enforce-download-offset-days: ${EFGS_ENFORCE_DOWNLOAD_OFFSET_DAYS:0}
     # The number of days to retain batch information in the database.
     retention-days: 14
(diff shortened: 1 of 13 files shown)


services/download/src/main/resources/application.yaml
```diff
@@ -14,6 +14,9 @@ services:
     # the keys for today. If it is set to 1, the keys from yesterday will be downloaded.
     # The value must be between 0 and 14.
     efgs-offset-days: ${EFGS_OFFSET_DAYS:1}
+    # The offset in days for which the keys from the federation gateway shall be downloaded from,
+    # ignoring if a batch of this day has already been downloaded.
+    efgs-enforce-download-offset-days: ${EFGS_ENFORCE_DOWNLOAD_OFFSET_DAYS:0}
```
EugenM-SAP, line 19: If i understand the parameter correctly, i suggest wording something like this : "This is the offset in past days for which keys should be downloaded again, regardless if a previous download had occurred. I.e. Value '3' means re-download all batches/keys from 3 days ago."

mibrasap: Thanks, updated it

services/download/src/test/resources/db/testdata-enforce-download-again/V99__createTestDataForDownloadTodayAgainIntegrationTest.sql
```diff
@@ -0,0 +1 @@
+INSERT INTO federation_batch_info(batch_tag, date, status) VALUES ('batch1_tag', current_timestamp, 'PROCESSED');
```
EugenM-SAP, line 1: Even though this approach of using flyway for initial test data is fine with me, there is an alternative to it. We can also use @Sql("/your_script.sql") annotation to load the sql statements needed for this test.  As you prefer.

EugenM-SAP (approved):

sorin-stefan-iovita-sap (approved):

EvgeniiSkrebtcov (approved):

ioangut (approved):

PR #899, opened 2020-10-20
Centralize logging for download/submission/upload/persistence modules

roesslerj:
## Description

This PR is according to EXPOSUREBACK-165: Refactor the way logging is done.  This offers following benefits

Easy maintenance of messages that gets logged in one central location.
Any future changes related to log messages would require change only in one file 
This file could be used as starting point in discussions with DPP colleagues
This would also come handy if we need to support multiple languages in future.

To achieve this, we centralized all log messages in one enum per module, all of which implement the same interface. By also implementing a wrapper for the slf4j logger, we achieve type-safety:

- app.coronawarn.server.common.Logger is the slf4j wrapper.
- app.coronawarn.server.common.LogMessages is the centralized interface.
- app.coronawarn.server.common.persistence.service.common.PersistenceLogMessages is the enum for the persistence module implementing the LogMessages interface.


mibrasap (changes requested):
Let's re-discuss the concept of this change. I see the advantages of having a centralized place for the log messages. The way it is done in this PR will also bring more complexity during development.

common/persistence/src/main/java/app/coronawarn/server/common/persistence/service/FederationBatchInfoService.java
```diff
@@ -44,7 +47,7 @@ public void updateStatus(FederationBatchInfo federationBatchInfo, FederationBatc
     String statusValue = status.name();
     federationBatchInfoRepository
         .saveDoUpdateStatusOnConflict(federationBatchInfo.getBatchTag(), federationBatchInfo.getDate(), statusValue);
-    logger.info("Marked batch {} with status {}.", federationBatchInfo.getBatchTag(), statusValue);
+    logger.info(MARKET_BATCH_WITH_STATUS, federationBatchInfo.getBatchTag(), statusValue);
```
kilbphilippSAP, line 50:
```suggestion
    logger.info(MARKED_BATCH_WITH_STATUS, federationBatchInfo.getBatchTag(), statusValue);
```

common/persistence/src/main/java/app/coronawarn/server/common/persistence/service/common/PersistenceLogMessages.java
```diff
@@ -0,0 +1,40 @@
...
+      "Deleting {} diagnosis key(s) with a submission timestamp older than {} day(s) ago."), //
+  DIAGNOSIS_KEYS_CONFLICTED_WITH_DB_ENTRIES(
+      "{} out of {} diagnosis keys conflicted with existing database entries and were ignored."), //
+  KEYS_PICKED_FROM_UPLOAD_TABLE("{} keys picked after read from upload table"), //
+  KEYS_REMAINING_AFTER_FILTERING_BY_CONSENT("{} keys remaining after filtering by consent"), //
+  KEYS_REMAINING_AFTER_FILTERING_BY_SHARE_POLICY("{} keys remaining after filtering by share policy"), //
+  KEYS_REMAINING_AFTER_FILTERING_BY_VALIDITY("{} keys remaining after filtering by validity"), //
+  KEYS_SELECTED_FOR_UPLOAD("Keys selected for upload: {}"), //
+  MARKET_BATCH_WITH_STATUS("Marked batch {} with status {}."), //
+  NR_RETRIEVED_DISCARDED_DIAGNOSIS_KEYS(
```
kilbphilippSAP, line 19: NUM / NO to abbreviate number in EN.

roesslerj:
As just discussed, we will drop this PR for the following reasons: Although centralizing logging statements makes it easier to gather them for DPP, it makes maintaining and adding them a lot more difficult. Using aspects seems to be counter-intuitive in this instance altogether.
Instead, we will create a mvn plugin to gather all logging statements in one file for DPP.

PR #903, opened 2020-10-20
Update artifact version

sorin-stefan-iovita-sap:


.mvn/maven.config
```diff
@@ -1,4 +1,4 @@
--Drevision=1.5.0-SNAPSHOT
+-Drevision=1.7.0-SNAPSHOT
```
mibrasap, line 1: Why do we already need 1.7?

mibrasap (changes requested):

.mvn/maven.config
EvgeniiSkrebtcov: should be 1.6 as we are preparing Beta 1.6 and 1.6 dev is not closed

sorin-stefan-iovita-sap: Done 😄

mibrasap (approved):

emmetsap (approved):

EvgeniiSkrebtcov (approved):

PR #933, opened 2020-10-26
Fix/duplicate efgs key distribution

KevponSAP:
## Checklist

## Description
Fixes an issue that keys are duplicated in different packages for EUR in some cases


services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/diagnosiskeys/DiagnosisKeyBundler.java
```diff
@@ -209,6 +209,10 @@ private void addKeyToMap(DiagnosisKey key, Map<String, List<DiagnosisKey>> keysB
             if (key.getOriginCountry().equals(originCountry) && !visitedCountry.equals(originCountry)) {
               return;
             }
+            if (!key.getOriginCountry().equals(originCountry) && !visitedCountry.equals(originCountry)
```
EugenM-SAP, line 212: for better maintainance it's worth to extract a method out of these 3 bool conditions. It's hard to understand why the key is skipped  just by looking at this code and not knowing the context.  Maybe even a javadoc to the new method which explains why the key is skipped or referneces the encountered bug.

mibrasap (approved):

EvgeniiSkrebtcov (approved):

hilmarf (approved):

PR #855, opened 2020-10-06
Remove Callback Service sources and references in Dockerfiles/POMs

JonasZeitzem-sap:
## Checklist

* [X] Test your changes as thoroughly as possible before you commit them. Preferably, automate your test by unit/integration tests.
* [X] Make sure `mvn install` runs for the whole project and, if you touched any code in the respective service, submission and distribution service can be run with `spring-boot:run`

## Description
Removes the Callback Service to get ready for 1.5 release. The Callbackservice is saved in a separate branch "with-callback-service" to proceed with development on the other branch.

PR #934, opened 2020-10-26
Revert d2917614 (#855): Remove Callback Service

roesslerj:
Revert d2917614 (#855): Remove Callback Service sources and references in Dockerfiles/POMs


EugenM-SAP:
Sonar coverage needs just a little bump to pass the 80% threshold. I suggest to focus on the APIErrorHandler unkown exception code path (by creating a test with a mock component that  throws a runtime exception from inside the callback functions and validate the 500 http return), rather than adding unit test for the Application class.

roesslerj:
Thanks for the hint, will do.

roesslerj:
We don't want to put this into 1.6 release.

PR #939, opened 2020-10-27
Merge changes from master to release/1.6

mibrasap:
Merge changes from master to release/1.6


services/upload/src/main/resources/application.yaml
```diff
@@ -25,6 +25,8 @@ services:
     min-batch-key-count: 140
     # The maximum number of keys that an upload batch can contain
     max-batch-key-count: 4000
+    # Time window Federation Upload Keys will be loaded from DB
+    retention-days: 14
```
hilmarf, line 29: we should  change this to 7 or even 6 to avoid issues

services/upload/src/main/java/app/coronawarn/server/services/federation/upload/keys/DiagnosisKeyPersistenceLoader.java
```diff
@@ -25,6 +25,8 @@ public DiagnosisKeyPersistenceLoader(FederationUploadKeyService uploadKeyService
   @Override
   public List<FederationUploadKey> loadDiagnosisKeys() {
     return this.uploadKeyService
-        .getPendingUploadKeys(ExpirationPolicy.of(uploadConfig.getExpiryPolicyMinutes(), ChronoUnit.MINUTES));
+        .getPendingUploadKeys(
+            ExpirationPolicy.of(uploadConfig.getExpiryPolicyMinutes(), ChronoUnit.MINUTES),
+            this.uploadConfig.getRetentionDays());
```
hilmarf, line 30: I don't get this...

services/upload/src/main/resources/application.yaml
hilmarf, line 29: 7

hilmarf (approved):

services/upload/src/main/java/app/coronawarn/server/services/federation/upload/keys/DiagnosisKeyPersistenceLoader.java
fredrb: What part? If it's the second parameter, it's a safety check to load keys only younger than 14 days. To avoid sending incorrect data to EFGS. 

Since the keys in the Upload table are duplicated form Diagnosis Key Table with a DB trigger, distribution will handle the retention policy. So this check shouldn't even be needed in the first place. It's just a sanity check to avoid any wrong keys leaking into our upload batch. Plus, in our test systems, old Test Data Generation keys _could_ end up with a wrong rolling start interval before being deleted.

services/upload/src/main/resources/application.yaml
fredrb: I think if we have anything smaller than 14 we risk losing keys, wouldn't we? 
The filter here is by rolling start interval - this means in a single submission we can have keys with rolling start interval from 14 days back if I'm not mistaken.

services/upload/src/main/java/app/coronawarn/server/services/federation/upload/keys/DiagnosisKeyPersistenceLoader.java
hilmarf: @fredrb but on which date is this checked?
1. submission timestamp?
1. rolling_start_interval_number?

PR #940, opened 2020-10-27
typo in comment

cougarten:
`[` should be `]`


cougarten:
my bad. I take it back, all correct.

PR #941, opened 2020-10-27
Better document CWA-hash vs ETag

roesslerj:
There was some confusion regarding the ETag vs. the CWA-hash header. This PR aims to improve the documentation and JavaDoc to fix this confusion. The calculation of the CWA-hash remains unchanged.


docs/DISTRIBUTION.md
```diff
@@ -113,15 +113,27 @@ AllUsers. This setting should only be used when running with the application wit
...
+
+```sh
+md5 -q export.bin | tr a-z A-Z | xxd -p -r | md5 -q
+```
+
+#### `ETag`
+
+[Standard HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag), used to determine whether the
+archive has changed and needs to be re-uploaded or re-downloaded by clients (i.e. mobile apps). Usually an MD5 of the
+complete archive file, but can be complex to calculate for multi-part uploads (i.e. see [this answer on StackOverflow](https://stackoverflow.com/questions/12186993/what-is-the-algorithm-to-compute-the-amazon-s3-etag-for-a-file-larger-than-5gb#answer-19896823).
```
mibrasap, line 135: There seems to be a missing word in this sentence

hilmarf (approved):

hilmarf (approved):

hilmarf (approved):

sorin-stefan-iovita-sap (approved):

PR #946, opened 2020-10-29
Dep/updates

hilmarf:
let's keep the dependency version management as easy as possible

with this changes, we just need to run from time to time:

`mvn versions:update-properties`

:-)


ioangut (approved):

sorin-stefan-iovita-sap (approved):
👍

PR #951, opened 2020-10-30
Default Values - Extend App Config with Configuration Parameters

sorin-stefan-iovita-sap:
## Description

This PR adds the required App Configuration Parameters default values described [here](https://github.com/corona-warn-app/cwa-app-tech-spec/blob/proposal/rsc-target-state/docs/spec/key-download.md#extend-app-config-with-configuration-parameters-for-key-download)


hilmarf (approved):

services/distribution/src/test/resources/application.yaml
```diff
@@ -65,6 +65,20 @@ services:
       min-ios: ${MIN_IOS_VERSION:0.5.0}
       latest-android: ${LATEST_IOS_VERSION:1.0.4}
       min-android: ${MIN_ANDROID_VERSION:1.0.4}
+    app-config-parameters:
+      ios-key-download-parameters:
+        cached-day-packages-to-update-on-etag-mismatch: ${KEY_DOWNLOAD_INVALIDATED_DAY_PACKAGES:[]}
```
mibrasap, line 70: For the test package resources.yaml, it might make sense not to rely on the environment variables but use the expected value only, e.g.:
cached-day-packages-to-update-on-etag-mismatch: []
This prevents failing tests due to a different system configuration

services/distribution/src/main/resources/application.yaml
```diff
@@ -95,6 +95,23 @@ services:
       min-ios: ${MIN_IOS_VERSION:0.5.0}
       latest-android: ${LATEST_IOS_VERSION:1.0.4}
       min-android: ${MIN_ANDROID_VERSION:1.0.4}
+    app-config-parameters:
+      ios-key-download-parameters:
+        cached-day-packages-to-update-on-etag-mismatch: ${KEY_DOWNLOAD_INVALIDATED_DAY_PACKAGES:"[]"}
```
mibrasap: Do we need the quotes in "[]" or can they be omitted, as it is the case in the test package's application.yaml?

services/distribution/src/test/resources/application.yaml
mibrasap, line 70: For the test package resources.yaml, it might make sense not to rely on the environment variables but use the expected value only, e.g.:
cached-day-packages-to-update-on-etag-mismatch: []
This prevents failing tests due to a different system configuration

mibrasap (changes requested):

services/distribution/src/main/resources/application.yaml
sorin-stefan-iovita-sap: They must be omitted because in the end the resulting string will contain the escaped quotes: i.e. "\"[]\""

mibrasap: Perfect :)

services/distribution/src/main/java/app/coronawarn/server/services/distribution/config/DistributionServiceConfig.java
```diff
@@ -654,4 +660,129 @@ public void setMinAndroid(String minAndroid) {
...
+      public String getCachedHourPackagesToUpdateOnETagMismatch() {
+        return cachedHourPackagesToUpdateOnETagMismatch;
+      }
+
+      public void setCachedHourPackagesToUpdateOnETagMismatch(String cachedHourPackagesToUpdateOnETagMismatch) {
+        this.cachedHourPackagesToUpdateOnETagMismatch = cachedHourPackagesToUpdateOnETagMismatch;
+      }
+    }
+
+    public static class IosKeyDownloadParameters extends CommonKeyDownloadParameters {
```
EugenM-SAP, line 748: Does IosKeyDownloadParameters have no specific params other than the common ones?

sorin-stefan-iovita-sap: Nope. Would you suggest to use only the common ones and remove the ios?

services/distribution/src/test/resources/application.yaml
sorin-stefan-iovita-sap: But then testing the default value mechanism would no longer be possible since the value [] is passed directly.

roesslerj (approved):

ioangut (dismissed):

services/distribution/src/main/java/app/coronawarn/server/services/distribution/config/DistributionServiceConfig.java
EugenM-SAP: Not necesarily. I think it's ok to leave it like this as it makes it explicit. I was just wondering whether some fields were forgotton

EugenM-SAP (approved):

services/distribution/src/test/resources/application.yaml
EugenM-SAP: But do we have a test where we don't use default ?

sorin-stefan-iovita-sap: Changes as suggested 😄

mibrasap (approved):

emmetsap (approved):

PR #955, opened 2020-11-02
Feature/download authentication exception

FelixRottler:
## Checklist

## Description

- shutdown of the download runner in case of a 403 response from the efgs


services/download/src/main/java/app/coronawarn/server/services/download/BatchDownloadException.java
```diff
@@ -0,0 +1,14 @@
+
+
```
hilmarf, line 2: please remove empty lines at the beginning of the file - the package declaration should be the first statement

hilmarf (changes requested):

roesslerj (approved):

hilmarf (approved):

PR #958, opened 2020-11-02
Add new Transmission risk encoding configuration in preparation of ENF v2 adoption

EugenM-SAP:
## Checklist

## Description

- This PR adds a configuration class that contains mappings which will be used to encode TRL in  DSOS and Report Type prior to distributing keys to the CDN. The intent is to allign with ENF v2 transmission risk calculations done on the client side. 
- The PR does not contain the actual invocation of these derivations (in a follow-up), only the mapping configurations.


services/distribution/src/main/resources/master-config/transmission-risk-encoding.yaml
```diff
@@ -0,0 +1,29 @@
...
+    8: 2
+  trl-to-rt:
+    1: 1
+    2: 2
+    3: 3
+    4: 4
+    5: 1
+    6: 2
+    7: 3
+    8: 4
```
hilmarf, line 29: @mlenkeit what do you think? All good?

hilmarf (approved):

services/distribution/src/main/resources/master-config/transmission-risk-encoding.yaml
mlenkeit: LGTM 👍

mlenkeit: Are there any automated tests to check if these parameters are actually pulled?

EugenM-SAP: There are tests to validate the values, and there will be a follow-up PR with the code that actually performs the encoding by using this map. There we will have tests which cover the scenario you mention

sorin-stefan-iovita-sap (approved):

PR #959, opened 2020-11-02
Bump Spring Boot version

sorin-stefan-iovita-sap:
## Description

This PR upgrades to the newest version of the Spring Boot. It also removes dependencies version that already come with this new version of the Spring Boot.


pom.xml
```diff
@@ -203,17 +195,6 @@
...
-        <groupId>junit</groupId>
-        <artifactId>junit</artifactId>
-        <version>${junit.version}</version>
-        <scope>test</scope>
-      </dependency>
-      <dependency>
-        <groupId>org.apache.tomcat.embed</groupId>
-        <artifactId>tomcat-embed-core</artifactId>
-        <version>${tomcat-embed-core.version}</version>
-      </dependency>
```
hilmarf, line 216: hmmm... shall we really remove that overriding? I have concerns, that the next security finding will occur in the same libraries... especially tomcat has very frequent new issues...

pom.xml
```diff
@@ -173,11 +170,6 @@
         <version>${guava.version}</version>
         <scope>runtime</scope>
       </dependency>
-      <dependency>
-        <groupId>org.apache.httpcomponents</groupId>
-        <artifactId>httpclient</artifactId>
-        <version>${httpclient.version}</version>
-      </dependency>
```
hilmarf, line 180: as long as it's the same version as spring uses it, we can keep it...

pom.xml
sorin-stefan-iovita-sap: Good point. I'll leave it as it was.

pom.xml
sorin-stefan-iovita-sap: I just reverted it. 😄

EugenM-SAP (approved):

hilmarf (approved):

PR #963, opened 2020-11-03
Use log4j %enc{%m} to encode user input in the logging

roesslerj:
As the documentation states: "This converter is particularly useful for encoding user provided data so that the output data is not written improperly or insecurely." Thanks to the input, messages are now also truncated after 1000 chars.

See https://logging.apache.org/log4j/2.x/manual/layouts.html for details


hilmarf (dismissed):
at least this removes control chars etc.
but it doesn't prevent large log sizes, right? What if someone sends 5GB of chars for countries? does log4j also have a MAX_SIZE or something similar for user input?

sorin-stefan-iovita-sap (approved):

roesslerj:
Adapted the PR to also take into account too big log messages, which are now truncated after 1000 chars (updated PR description).

hilmarf (approved):

kilbphilippSAP (approved):

EugenM-SAP (approved):

PR #965, opened 2020-11-05
Remove disable csrf protection for callback get request

ioangut:
## Checklist

## Description

Enable csrf protection for callback service


sorin-stefan-iovita-sap (approved):

hilmarf (approved):

hilmarf:
is that all we need? just disabling it? what about adjusting integration tests? what needs to be done then on client side (EFGS)?

PR #972, opened 2020-11-09
Fix Infinite processing of the same batch tag

### LINKED ISSUE #970: [BSI][20201107] Denial-of-Service in Download Service
Rating: Informational

Description:
The download service component of the cwa-server backend can be forced into an endless loop by a malicious or faulty federation gateway server. When downloading new batches from the federation gateway server, the download service checks the nextBatchTag header of the federation gateway server’s response. If this header is set, the download service will continue requesting a batch with this tag. It will not stop as long as the server includes this header. This can trap the download service in an endless loop.
This highly depends on the behavior and implementation of the federation gateway server. However, even a non-malicious, faulty implementation might trigger this issue. Additionally, the issue was sometimes hard to reproduce as the server first has to be forced to request new batches (which in some cases required setting the EFGS_ENABLE_DATE_BASED_DOWNLOAD configuration to true).

Affected System:
cwa-server (commit 4f3f460f580ea5d548dff0e3ceb739908c83d1a7)

Proof of Concept:
The following screenshots show the offending lines in the FederationBatchProcessor.java:

Download Service DoS Offending Code

The following snippets show a minimalistic federation gateway server mock that returns invalid data but sets the nextBatchTag header on each response. The first request needs to fail (in this case implemented by sending an invalid HTTP request body) so that the batch tag is added to the list of unprocessed batches.
snippet-1
snippet-2

The following screenshot shows the failing first request on the first run of the download service:
Download Service DoS First Invalid Request

The next screenshot shows the log output of the download service when it is run again. It can be seen that the server will endlessly try to request a next batch.
Download Service DoS Endless Loop Log

Recommended Controls: 
The while loop should be adapted so that it is not possible for it to continue forever. The received nextBatchTag header should be checked. If it is already contained in the list of unprocessed batches, it should not be added again. This does not protect against a malicious server that sends different batch tags on each request. Therefore, alternatively, a timeout value or a maximum number of retries could be introduced.

sorin-iovita:
## Description

This PR fixes the infinite processing of the same batch tag id.
closes #970


services/download/src/main/java/app/coronawarn/server/services/download/FederationBatchProcessor.java
```diff
@@ -25,7 +25,9 @@
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Optional;
+import java.util.Queue;
 import java.util.concurrent.atomic.AtomicBoolean;
+import org.apache.commons.collections4.queue.CircularFifoQueue;
```
hilmarf, line 30: in my opinion a bit over engineered... why isn't `java.util.Set` / HashSet sufficient?

services/download/pom.xml
```diff
@@ -20,6 +23,11 @@
       <artifactId>postgresql</artifactId>
       <scope>test</scope>
     </dependency>
+      <dependency>
+          <groupId>org.apache.commons</groupId>
+          <artifactId>commons-collections4</artifactId>
+          <version>${commons-collections4.version}</version>
+      </dependency>
```
hilmarf, line 30: I would like to not depend on more and more other libraries...

services/download/src/main/resources/application.yaml
```diff
@@ -19,6 +19,8 @@ services:
     retention-days: 14
     validation:
       allowed-report-types: ${ALLOWED_REPORT_TYPES:CONFIRMED_TEST,CONFIRMED_CLINICAL_DIAGNOSIS}
+    # The number of batch tags that will be cached in memory
+    cached-batches-size: ${CACHED_BATCHES_SIZE:1000}
```
hilmarf, line 23: MAX_BATCH_COUNT

hilmarf (changes requested):

services/download/src/main/java/app/coronawarn/server/services/download/FederationBatchProcessor.java
```diff
@@ -130,15 +133,21 @@ public void processUnprocessedFederationBatches() throws FatalFederationGatewayE
 
     while (!unprocessedBatches.isEmpty()) {
       FederationBatchInfo currentBatch = unprocessedBatches.remove();
+      cachedBatches.add(currentBatch);
       processBatchAndReturnNextBatchId(currentBatch, ERROR)
           .ifPresent(nextBatchTag -> {
-            if (config.getEfgsEnforceDateBasedDownload()) {
-              unprocessedBatches.add(new FederationBatchInfo(nextBatchTag, currentBatch.getDate()));
+            FederationBatchInfo batch = new FederationBatchInfo(nextBatchTag, currentBatch.getDate());
```
mibrasap, line 139: For clarity: please indicate in the variable name that it is only a batch info, but not the batch itself

services/download/src/main/java/app/coronawarn/server/services/download/FederationBatchProcessor.java
```diff
@@ -42,6 +43,7 @@
   private final FederationGatewayDownloadService federationGatewayDownloadService;
   private final DownloadServiceConfig config;
   private final ValidFederationKeyFilter validFederationKeyFilter;
+  private final FifoMaxEntriesSet<FederationBatchInfo> cachedBatches;
```
mibrasap, line 46: Did you check if there is a simpler solution for that? What about a list that stores all the processed batch tags and compare incoming "next batch tags" against this list?

services/download/src/test/java/app/coronawarn/server/services/download/FederationBatchProcessorTest.java
```diff
@@ -226,6 +228,24 @@ void testOneUnprocessedBatchFails() throws Exception {
...
+
+      when(batchInfoService.findByStatus(UNPROCESSED)).thenReturn(list(batchInfo));
+      when(federationGatewayDownloadService.downloadBatch(batchTag1, date)).thenReturn(serverResponse);
+
+      Assertions
+          .assertTimeoutPreemptively(Duration.ofSeconds(1), () -> batchProcessor.processUnprocessedFederationBatches());
+      Mockito.verify(batchInfoService, times(1)).findByStatus(UNPROCESSED);
+      Mockito.verify(federationGatewayDownloadService, times(1)).downloadBatch(batchTag1, date);
+      Mockito.verify(batchInfoService, times(1)).updateStatus(batchInfo, PROCESSED);
+
```
mibrasap, line 247: unnecessary empty line

mibrasap (changes requested):

services/download/src/main/java/app/coronawarn/server/services/download/FederationBatchProcessor.java
```diff
@@ -130,10 +134,14 @@ public void processUnprocessedFederationBatches() throws FatalFederationGatewayE
...
       FederationBatchInfo currentBatch = unprocessedBatches.remove();
+      cachedBatches.add(currentBatch);
       processBatchAndReturnNextBatchId(currentBatch, ERROR)
           .ifPresent(nextBatchTag -> {
+            FederationBatchInfo batch = new FederationBatchInfo(nextBatchTag, currentBatch.getDate());
             if (config.getEfgsEnforceDateBasedDownload()) {
-              unprocessedBatches.add(new FederationBatchInfo(nextBatchTag, currentBatch.getDate()));
+              if (!cachedBatches.contains(batch)) {
+                unprocessedBatches.add(batch);
+              }
```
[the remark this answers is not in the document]
mibrasap: Do we really want to introduce a maximum number of batches that can be downloaded?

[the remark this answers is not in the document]
roesslerj: I am also against this. I re-read the original ticket "This does not protect against a malicious server that sends different batch tags on each request." and I think we should have other means of guarding against a malicious servers in the backend (e.g. SSL etc.). We don't know how many cases there will be in Germany and Europe, so limiting the number of batches seems dangerous...

services/download/src/main/java/app/coronawarn/server/services/download/FederationBatchProcessor.java
roesslerj: Removed the dependency and added a FIFO Queue with HashSet backing.

services/download/pom.xml
roesslerj: Removed dependency.

services/download/src/main/resources/application.yaml
roesslerj: Renamed it to "MAX_BATCH_LOOP_GUARD" as I think this better describes its purpose. But very open here :-)

services/download/src/main/java/app/coronawarn/server/services/download/FifoMaxEntriesSet.java
```diff
@@ -0,0 +1,56 @@
...
+   * @param e The element to add
+   * @return {@code true} if this set did not already contain the specified
+   *         element
+   */
+  public boolean add(E e) {
+    fast.add(e);
+    order.add(e);
+    while (fast.size() > maxSize) {
+      fast.remove(order.remove());
+    }
```
hilmarf, line 39: @sorin-stefan-iovita-sap / @roesslerj 
sorry, but I still don't get it... let's assume the following: we have max size of 3 and batch-IDs A, B, C, D, A, B, C, D, ...
when `fast` reaches `maxSize`, it contains A,B,C,D - it will remove A as it is the first element in `order`...
So with next batch being again ID: A, we are stuck in a loop again, aren't we?

hilmarf (changes requested):

services/download/src/main/java/app/coronawarn/server/services/download/FederationBatchProcessor.java
roesslerj: Imho we need to limit the amount of batches we lookup, or else we're prone to a memory leak. Therefore a simple list doesn't suffice. But again—very open to different solution proposals :-)

services/download/src/main/java/app/coronawarn/server/services/download/FifoMaxEntriesSet.java
roesslerj: Yes. But there is a trade-off: Either no maxSize and possibly introduce memory leak—or have maxSize and not detect loops that are bigger than maxSize. I am ok with either, we just can't have both and need to decide what we want...

services/download/src/main/java/app/coronawarn/server/services/download/FederationBatchProcessor.java
roesslerj: Done. Also in the line above (nextBatchTag).

services/download/src/main/java/app/coronawarn/server/services/download/FederationBatchProcessor.java
roesslerj: Will just use a HashSet.

services/download/src/test/java/app/coronawarn/server/services/download/FederationBatchProcessorTest.java
roesslerj: Removed.

services/download/src/main/java/app/coronawarn/server/services/download/FifoMaxEntriesSet.java
roesslerj: Resolved as discussed :-)

services/download/src/main/java/app/coronawarn/server/services/download/FederationBatchProcessor.java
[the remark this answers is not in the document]
roesslerj: As discussed: We'll run into OutOfMemory if there are too many batches, rather than setting a fixed limit...

services/download/src/main/java/app/coronawarn/server/services/download/FederationBatchProcessor.java
mibrasap: Nice, a set makes indeed more sense than a list in this case :)

mibrasap (approved):

hilmarf (approved):

PR #974, opened 2020-11-10
Improve readability by using %maxLen{%m}{20} instead of %.-1000m

roesslerj:
Very small change that will improve the readability of the logging config.


hilmarf (approved):
what about `upload` service?

sorin-iovita (approved):

PR #991, opened 2020-11-18
Rel/1.7

mibrasap:
## Checklist


sorin-iovita (approved):

services/callback/src/main/resources/log4j2.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<Configuration status="WARN">
+<Configuration status="WARN" shutdownHook="disable">
```
hilmarf, line 2: @roesslerj - what does this disabling of the shutdown hook do? #589

services/upload/src/main/resources/log4j2-dev.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration status="WARN">
```
hilmarf, line 2: @roesslerj - what about keeping everything in sync: ` shutdownHook="disable" `?

hilmarf: forget about it... didn't realize it's DEV only

PR #1007, opened 2020-11-24
App Statistics File

fredrb:
- Introduces a new file `stat` during the Distribution
- Introduces new protobuf spec for Statistic Object
- Assigns data points to the respective Key Figure Cards according to [spec](https://github.com/corona-warn-app/cwa-app-tech-spec/blob/1562caab2bac7f911556ee5ed7315c32332f192a/docs/spec/statistics.md)
- Loads test file from filesystem

What is pending on this feature is to load the actual files from TSI. For the time being, we are loading a fake statistics file from filesystem.


services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/component/StatisticsStructureProvider.java
```diff
@@ -0,0 +1,41 @@
...
+import app.coronawarn.server.services.distribution.assembly.structure.Writable;
+import app.coronawarn.server.services.distribution.assembly.structure.WritableOnDisk;
+import app.coronawarn.server.services.distribution.config.DistributionServiceConfig;
+import app.coronawarn.server.services.distribution.statistics.directory.StatisticsDirectory;
+import org.springframework.stereotype.Component;
+
+
+/**
+ * Reads configuration parameters from the respective files in the class path and builds a {@link
+ * StatisticsStructureProvider} with them.
```
EugenM-SAP, line 13: Did you mean build. statistics archive with them?

services/distribution/src/main/java/app/coronawarn/server/services/distribution/objectstore/client/StatsS3ReadingConfig.java
```diff
@@ -0,0 +1,40 @@
...
+    return createClient(distributionServiceConfig.getStatistics());
+  }
+
+  private ObjectStoreClient createClient(StatisticsConfig s3stats) {
+    AwsCredentialsProvider credentialsProvider = StaticCredentialsProvider.create(
+        AwsBasicCredentials.create(s3stats.getAccessKey(), s3stats.getSecretKey()));
+    String endpoint = removeTrailingSlash(s3stats.getEndpoint());
+
+    return new S3ClientWrapper(S3Client.builder()
+        .region(Region.of("region-1"))
```
EugenM-SAP, line 30: Should this be a constant. I see in other configs we have: 
private static final Region DEFAULT_REGION = Region.EU_CENTRAL_1;

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/StatisticsToProtobufMapping.java
```diff
@@ -0,0 +1,133 @@
...
+          try {
+            card = keyFigureCardFactory.createKeyFigureCard(stat, id);
+            figureCardMap.put(id, card);
+          } catch (MissingPropertyException ex) {
+            logger.warn(ex.getMessage());
+          }
+        }
+      });
+
+      if (figureCardMap.values().stream().allMatch(Objects::nonNull)) {
```
EugenM-SAP, line 121: What is the purpose of checking whether values are null after the perivous loop? If no card was created with the previous stat than it should stop? Should we throw an exception ?

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/file/RemoteStatisticJsonFileLoader.java
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.context.annotation.Profile;
+import org.springframework.stereotype.Component;
+
+@Component
+@Profile("!local-json-stats")
+public class RemoteStatisticJsonFileLoader implements JsonFileLoader {
+
+  @Autowired
+  @Qualifier("stats-s3")
+  ObjectStoreClient s3Stats;
```
EugenM-SAP, line 16: fields should  be private and injected via constructor for consistency with the rest of the code

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/keyfigurecard/factory/EmptyCardFactory.java
```diff
@@ -0,0 +1,33 @@
...
+    return KeyFigureCardSequenceConstants.FOURTH_CARD_ID;
+  }
+
+  @Override
+  protected KeyFigureCard buildKeyFigureCard(StatisticsJsonStringObject stats, Builder keyFigureBuilder) {
+    return keyFigureBuilder.build();
+  }
+
+  @Override
+  protected List<Pair<String, Optional<Object>>> getNonNullFields(StatisticsJsonStringObject stats) {
```
EugenM-SAP, line 30: Just a suggestion. If applicable use Map instead of List of Pairs

services/distribution/src/main/resources/application.yaml
```diff
@@ -115,7 +126,8 @@ services:
         max-exposure-detections-per-interval: ${ANDROID_EXPOSURE_DETECTION_MAX_ED_PER_INTERVAL:6}
         overall-timeout-in-seconds: ${ANDROID_EXPOSURE_DETECTION_OVERALL_TIMEOUT_IN_SECONDS:900}
 
-
+remote-file-client:
+  base-url: 'https://obs.eu-de.otc.t-systems.com/obs-cwa-public-dev'
```
EugenM-SAP, line 130: Is there an impact to using this URL when the app is deployed in differnet environments (INT, DEV, WRU ) ?

services/distribution/src/test/java/app/coronawarn/server/services/distribution/objectstore/integration/ObjectStoreFilePreservationIT.java
```diff
@@ -79,20 +79,19 @@ public void setup() throws IOException {
    * The test presumes there are 4 consecutive days with 80 keys submitted daily. Running a distribution in Day 4 would
    * result in:
    * <p>
-   *     Day 1   -> submission of 80 keys   -> 1 empty file distributed<br>
-   *     Day 2   -> submission of 80 keys   -> 1 distributed containing 160 keys<br>
-   *     Day 3   -> submission of 80 keys   -> 1 empty file distributed<br>
-   *     Day 4   -> submission of 80 keys   -> 1 distributed containing 160 keys<br>
+   * Day 1   -> submission of 80 keys   -> 1 empty file distributed<br> Day 2   -> submission of 80 keys   -> 1
```
EugenM-SAP, line 82: I think formating this class produces a weird javadoc. Can we revert it ?

services/submission/src/main/resources/application.yaml
```diff
@@ -87,7 +87,7 @@ management:
     web:
       exposure:
         include: 'health, prometheus'
-  metrics:
+  metrics:\
```
EugenM-SAP, line 90: is the backslash needed?

roesslerj (changes requested):
Other than those small things, this looks good :-)

services/distribution/src/main/java/app/coronawarn/server/services/distribution/objectstore/client/StatsS3ReadingConfig.java
fredrb: Agreed. We can create a new constant for it. It's not `Region.EU_CENTRAL_1` though. It's really a region string of `region-1`...

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/StatisticsToProtobufMapping.java
ioangut: The list of statistics is ordered by date, and if those 4 cards(INFECTIONS_CARD, INCIDENCE_CARD, KEY_SUBMISSION,FOURTH_CARD) are already populated with values from the most recent statistics entries from the json file, then we should not look into the older ones, so we stop the loop.

fredrb: It's checking for `nonNull`. The loop is going through the most recent record until the oldest, as Ioan mentioned. If half-way through we already have all four cards created, we can skip break out of the loop and return all `KeyFigureCard` objects.

services/distribution/src/main/resources/application.yaml
ioangut: I removed this since it was used for local testing, and replaced with a different approach

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/StatisticsToProtobufMapping.java
EugenM-SAP: Sorry, my bad. I misread isNull instead of nonNull

EugenM-SAP (dismissed):

roesslerj (dismissed):

EugenM-SAP (dismissed):

codestyle/checkstyle.xml
```diff
@@ -60,7 +60,6 @@
             <property name="allowByTailComment" value="true"/>
             <property name="allowNonPrintableEscapes" value="true"/>
         </module>
-        <module name="AvoidStarImport"/>
```
roesslerj, line 63: Do we really want to change the checkstyle definition with this PR? Also, this rule seems to be an easy one to meet...

EugenM-SAP (approved):

codestyle/checkstyle.xml
roesslerj: This change is only needed for two instances of `KeyFigureCardSequenceConstants.*`...

roesslerj (changes requested):

roesslerj (approved):

PR #1023, opened 2020-11-30
Merge/1.7 into master

hilmarf:
see #1021


services/distribution/src/main/resources/master-config/v2/risk-calculation-parameters.yaml
```diff
@@ -13,6 +13,7 @@ minutes-at-attenuation-filters:
       min: 0
       max: 73
       max-exclusive: true
+  -
```
hilmarf, line 16: @mlenkeit this additional `-` is correct, right?

sorin-iovita (dismissed):

services/distribution/src/main/resources/master-config/v2/risk-calculation-parameters.yaml
mlenkeit: no, thanks for catching this!

mlenkeit (changes requested):
see https://github.com/corona-warn-app/cwa-server/pull/1023/files#r532390312

mlenkeit (approved):
LGTM

fredrb (approved):

sorin-iovita (approved):

PR #1058, opened 2020-12-09
Make statistics module more resilient

fredrb:
This PR introduces a mechanism to handle exceptions on S3 client more gracefully. 

1. Retries S3 request in case fails (max 3) 
2. Logs message with bucket name in case bucket not found
3. Logs message with file path in case unauthorized or file not found
4. Logs generic connection message in case connection failed
5. Distribution will not stop if statistics object store is not available 

**Internal Ticket:** EXPOSUREBACK-647


services/distribution/src/main/java/app/coronawarn/server/services/distribution/objectstore/client/S3ClientWrapper.java
```diff
@@ -62,6 +62,10 @@ public boolean bucketExists(String bucketName) {
   }
 
   @Override
+  @Retryable(
```
hilmarf, line 65: Do we use this `@Retryable` in upload/download as well? If not, why?

hilmarf (approved):

sorin-iovita (approved):

services/distribution/src/main/java/app/coronawarn/server/services/distribution/objectstore/client/S3ClientWrapper.java
fredrb: We don't. And I don't know why. 🙂

PR #1063, opened 2020-12-10
Feature/rename callback truststore

ioangut:
## Checklist

* [x] Make sure `mvn -P integration-tests clean verify` runs for the whole project and, if you touched any code in the respective [service](services), ensure it can be run with `spring-boot:run`

## Description

This PR modifies the names of the Env variables for Federation Integration services.


.circleci/config.yml
```diff
@@ -1,54 +0,0 @@
-version: 2.1
```
roesslerj, line 1: Why do you delete all of those files? I would prefer to have this in a separate PR, named accordingly...

roesslerj (changes requested):

.circleci/config.yml
ioangut: We are looking into this, since it was not intended to do it.

PR #1069, opened 2020-12-11
Hf/1.8.1

hilmarf:
@tklingbeil please have a close look!


mlenkeit (approved):
Multiple spot checks all looked good to me. Also checked the general pattern and the highest TRL is always before the midpoint of the interval, while it was the other way before.

mibrasap (approved):

PR #1077, opened 2020-12-15
Enhance Upload Service to support multiple destinations - db changes

EugenM-SAP:
## Checklist

* [ ] Make sure `mvn -P integration-tests clean verify` runs for the whole project and, if you touched any code in the respective [service](services), ensure it can be run with `spring-boot:run`

## Description

Introduces enhancements for the Upload Service with regards to integration with the Swiss Federation Gateway (SGS):
- New dedicated table to store keys which are uploadable to the SGS (name: 'sgs_upload_key')
- New db user and role for the sgs_upload_key table. This user will be assigned to the Upload job instance that is running.
- New DB triggers to replicate diagnosis keys to sgs_upload_key as well as to remove keys from it when retention job is running.
- Extend current integration test to cover replication and retentionof keys for  both federation gateway tables.

Changes to parameterize the Upload Service and make the target configurable will be provided in a follow-up PR


roesslerj:
Can you notify once the build works?

PR #1086, opened 2020-12-16
Enhance Upload Service to support multiple destinations

EugenM-SAP:
## Checklist

* [x] Make sure `mvn -P integration-tests clean verify` runs for the whole project and, if you touched any code in the respective [service](services), ensure it can be run with `spring-boot:run`

## Description

Introduces enhancements for the Upload Service with regards to integration with the Swiss Federation Gateway (SGS):

*New dedicated table to store keys which are uploadable to the SGS (name: 'sgs_upload_key')
* New db user and role for the sgs_upload_key table. This user will be assigned to the Upload job instance that is running.
* New DB triggers to replicate diagnosis keys to sgs_upload_key as well as to remove keys from it when retention job is running.
* Extend current integration test to cover replication and retention of keys for both federation gateway tables.
* Add separate profiles for connection to SGS and EFGS ("connect-efgs", "connect-sgs"). The service has to retrieve data from different tables depending on the target system it has to upload it to. There were two aproaches considered for this purpose:
     * Parameterize the upload key queries with a dynamic tablename. This approach however has the downside of migrating the data access specific code to using JdbcTemplate (or other classic methods) instead of the current Spring Jdbc Repositories layer. This is because @Query does not support dynamic table parametrization (via SpEL for example).
     * Create separate profiles with separate repository implementations. This is the approach taken in this PR and although it requires duplication of @Query statements for each context, it decouples the SGS and EFGS connection completely which can help later on if their requirements around data model evolve separately.


setup/setup-roles.sql
```diff
@@ -76,3 +76,12 @@ CREATE ROLE cwa_federation_download
    NOCREATEROLE
    NOREPLICATION
    IN ROLE cwa_user;
+
```
FelixRottler, line 79: I assume that this user is only for the sgs table. I think it would also make sense to have AC here. So that this swiss Role cannot have access to the other efgs table. What is your opinion on this?

FelixRottler (approved):

ioangut (approved):

PR #1112, opened 2021-01-06
Updated DSOS-->TRL as per Thomas Klingbeil from 2021-01-06

hilmarf:
One more thing (relevant for app side): The RKI colleagues asked us to change the meaning for the “zero point” (e.g. 700) once more: The zero point so far marked the youngest day within the range. The requested change is that we shift the number by 1 towards “now”, i.e.:

For a 7-day range (“within the last week”), 700 would then be today (date of submission) etc.
For a 7-day range (“1-2 weeks ago”), 700 would then be exactly one week before the submission and the day of submission would be 707 (meaning, we’d go back until 693)
For a 7-day range (“more than two weeks ago”), 700 would be two weeks before submission, and the day of submission would be 714.


mlenkeit (changes requested):
Looks like lots of mappings have been dropped. I don't think that's correct, will reach out to you directly to discuss...

common/persistence/src/main/resources/derivation-maps.yaml
```diff
@@ -274,6 +272,7 @@ tek-field-derivations:
     1797: 2
     1798: 2
     1799: 2
+    1800: 1
```
hilmarf, line 275: @mlenkeit what do you think, shall we remove the mappings to 1, since it's the default

mlenkeit (approved):
LGTM now, at least as far as the consistency with the spreadsheet is concerned.

mibrasap (approved):

PR #1114, opened 2021-01-06
Updated DSOS-->TRL as per Thomas Klingbeil from 2021-01-06

hilmarf:


mlenkeit (changes requested):
Looks like lots of mappings have been dropped. I don't think that's correct, will reach out to you directly to discuss...

mlenkeit (approved):
LGTM now, at least as far as the consistency with the spreadsheet is concerned.

mibrasap (approved):

PR #1151, opened 2021-01-18
Gh pages additional releases

emmetsap:
Additional Javadoc's for subsequent releases since 1.0


hilmarf (approved):

hilmarf:
Hi @emmetsap, can you please adjust also the index.html, so that the additional versions are listed here: https://corona-warn-app.github.io/cwa-server/

PR #1157, opened 2021-01-21
add extra tests to JSON validation on statistic module

fredrb:
Adds extra test automation to how Statistics Module handles the external JSON:
- If JSON has objects that don't mean anything: shouldn't crash, generates empty cards
- Parsing should filter out based on effective date 
- Log level should be `error` if statistic fails

### DIFF
--- a/services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/StatisticsToProtobufMapping.java
+++ b/services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/StatisticsToProtobufMapping.java
@@ -74,8 +74,8 @@ public class StatisticsToProtobufMapping {
           .addAllKeyFigureCards(buildAllKeyFigureCards(jsonStringObjects))
           .build();
     } catch (BucketNotFoundException | ConnectionException | FilePathNotFoundException ex) {
-      logger.warn(ex.getMessage());
-      logger.warn("Statistics file will not be generated due to previous errors!");
+      logger.error(ex.getMessage());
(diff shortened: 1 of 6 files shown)


services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/StatisticsToProtobufMapping.java
```diff
@@ -74,8 +74,8 @@ public Statistics constructProtobufStatistics() {
           .addAllKeyFigureCards(buildAllKeyFigureCards(jsonStringObjects))
           .build();
     } catch (BucketNotFoundException | ConnectionException | FilePathNotFoundException ex) {
-      logger.warn(ex.getMessage());
-      logger.warn("Statistics file will not be generated due to previous errors!");
+      logger.error(ex.getMessage());
+      logger.error("Statistics file will not be generated due to previous errors!");
```
hilmarf, line 78: I would prefer to keep it in one single line... loggers behave from time to time differently and not all lines are kept always together - don't mentioning log analyzer tools...

```suggestion
      logger.error("{} - Statistics file will not be generated due to previous errors!", ex.getMessage());
```
or
```suggestion
      logger.error("Statistics file will not be generated due to previous errors!", ex);
```

hilmarf (changes requested):

roesslerj (dismissed):

hilmarf (approved):

roesslerj (approved):

Issue #318, opened 2020-05-26
Consider Basic Auth for Actuator endpoint


johanneseschrig:
Communicate with Infrastructure, if and how we should enable basic auth on the submission actuator endpoint.

roesslerj:
Closing this, as it's not relevant anymore...

PR #1172, opened 2021-02-02
liveness & readiness probe for callback service

kilbphilippSAP:
## Checklist

## Description

Enable liveness and readiness probe for callback service on port 8081

### DIFF
--- a/services/callback/src/main/java/app/coronawarn/server/services/callback/config/SecurityConfig.java
+++ b/services/callback/src/main/java/app/coronawarn/server/services/callback/config/SecurityConfig.java
@@ -4,7 +4,7 @@ import static java.util.Collections.emptyList;
 
 import app.coronawarn.server.services.callback.CertificateCnMismatchException;
 import app.coronawarn.server.services.callback.controller.CallbackController;
-import java.util.Arrays;
+import java.util.Collections;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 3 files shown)


services/callback/src/main/resources/application.yaml
```diff
@@ -53,7 +55,7 @@ management:
     health:
       group:
         readiness:
-          include: db, verificationService
+          include: db, callbackController
```
hilmarf, line 58: :astonished: `verificationService` ???

hilmarf (dismissed):

hilmarf:
hey @kilbphilippSAP if possible, can you also check on issue #318 - which is kind of related... but before we're enforcing this, we need to check with operations, if that's possible for their tools.

services/callback/src/main/java/app/coronawarn/server/services/callback/config/SecurityConfig.java
```diff
@@ -28,8 +28,13 @@
 
   private static final String CALLBACK_ROUTE =
       "/version/v1" + CallbackController.CALLBACK_ROUTE;
-  private Logger logger = LoggerFactory.getLogger(SecurityConfig.class);
-  private CallbackServiceConfig callbackServiceConfig;
+  private static final String ACTUATOR_ROUTE = "/actuator/";
+  private static final String HEALTH_ROUTE = ACTUATOR_ROUTE + "health";
+  private static final String PROMETHEUS_ROUTE = ACTUATOR_ROUTE + "prometheus";
+  private static final String READINESS_ROUTE = ACTUATOR_ROUTE + "readiness";
```
FelixRottler, line 34: Can we verify that this URL is correct? I tried it locally and since readiness is custom group of health it should be under health/readiness?
```suggestion
  private static final String READINESS_ROUTE = HEALTH_ROUTE + "/readiness";
```

services/callback/src/main/java/app/coronawarn/server/services/callback/config/SecurityConfig.java
```diff
@@ -28,8 +28,13 @@
...
+  private static final String LIVENESS_ROUTE = ACTUATOR_ROUTE + "liveness";
```
FelixRottler, line 35: Same here

services/callback/src/main/resources/application.yaml
FelixRottler: As far as I understand this "callbackController" should correspond to a @Component("callbackController) which should inherit from HealthIndicator, which we do not have or? If yes then this can be removed

FelixRottler (changes requested):

FelixRottler (approved):

hilmarf (approved):

PR #1175, opened 2021-02-03
chore(app-config): increase upper bound for high risk per day

mlenkeit:
This is in response to EXPOSUREAPP-4856, which was found internally during testing. The tester hat 1.687 exposure windows on the device accumulating to a normalized time per day of more than 11.000 minutes (aka _risk minutes_).

This PR changes the upper bound by accomodate this scenario. We have no indication that this happens in reality but who knows...


tklingbeil (approved):

tklingbeil:
Works for me. This should now equal to permanent (!) close (!) contact to 2000 people per day...

hilmarf (approved):

PR #1181, opened 2021-02-04
Feature/download resilience

mibrasap:


services/download/src/main/java/app/coronawarn/server/services/download/FederationBatchProcessor.java
```diff
@@ -157,12 +158,14 @@ private boolean isEfgsEnforceDateBasedDownloadAndNotSeen(String batchTag) {
     LocalDate date = batchInfo.getDate();
     String batchTag = batchInfo.getBatchTag();
     logger.info("Processing batch for date {} and batchTag {}.", date, batchTag);
+    AtomicReference<Optional<String>> nextBatchTag = new AtomicReference<>(Optional.empty());
     try {
       BatchDownloadResponse response = federationGatewayDownloadService.downloadBatch(batchTag, date);
+      AtomicBoolean batchContainsInvalidKeys = new AtomicBoolean(false);
       if (config.isBatchAuditEnabled()) {
         federationGatewayDownloadService.auditBatch(batchTag, date);
```
hilmarf, line 166: what happens, if the auditing fails? is an exception thrown? then the nextBatchTag won't be set, right?

services/download/src/main/java/app/coronawarn/server/services/download/FederationBatchProcessor.java
```diff
@@ -157,12 +158,14 @@ private boolean isEfgsEnforceDateBasedDownloadAndNotSeen(String batchTag) {
...
       }
-      AtomicBoolean batchContainsInvalidKeys = new AtomicBoolean(false);
+      nextBatchTag.set(response.getNextBatchTag());
```
hilmarf, line 168: I guess this have to move before the audit happens

hilmarf (changes requested):
move audit a few lines up

services/download/src/main/java/app/coronawarn/server/services/download/FederationBatchProcessor.java
mibrasap: This has become obsolete, since the audit call was moved down with another PR

hilmarf (approved):

EugenM-SAP (approved):

PR #1207, opened 2021-02-19
HF1.12.1 - drop-if-minutes-in-range, max: 5

hilmarf:
to update the mobile app-config


mlenkeit:
We confirmed that this is the correct change. But let's hold back the PR until everything else is finalized

mibrasap (approved):

PR #1231, opened 2021-02-26
Feature/v2 app configuration

ioangut:
## Checklist

## Description

Host additional App Config on CDN for version 2 of the app_config_ios and app_config_android files.


services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/appconfig/ApplicationConfigurationV2PublicationConfig.java
```diff
@@ -38,35 +43,77 @@
...
  */
 @Configuration
 public class ApplicationConfigurationV2PublicationConfig {
 
   /**
    * The location of the exposure configuration source files for Android and Ios.
    */
-  private static final String V2_RISK_PARAMETERS_FILE = "main-config/v2/risk-calculation-parameters.yaml";
+  private static final String V1_RISK_PARAMETERS_FILE = "main-config/v2/risk-calculation-parameters.yaml";
+  private static final String V2_RISK_PARAMETERS_FILE = "main-config/v2/risk-calculation-parameters-v2.yaml";
```
mlenkeit, line 64: I find it a bit confusing that the path contains `v2` twice (although I understand why). How about changing the filename to `risk-calculation-parameters-1.15.yaml`? Or we change the `v2` in between to `v3` and host the app config on CDN also under `v3` instead of `v2`?

ioangut: I will change the filename to risk-calculation-parameters-1.15.yaml

mlenkeit (approved):
LGTM as far as I can tell 👍

mlenkeit:
It seems Sonar is failing because the coverage is too low. Is it because I removed some redundant config in https://github.com/corona-warn-app/cwa-server/pull/1231/commits/c6ec2a1114d7424bfdaa5e7aab69cccf6682e300?

ioangut:
> It seems Sonar is failing because the coverage is too low. Is it because I removed some redundant config in [c6ec2a1](https://github.com/corona-warn-app/cwa-server/commit/c6ec2a1114d7424bfdaa5e7aab69cccf6682e300)?
Sonar cloud is failing since the coverage on this branch was always too low, but for the first builds, for some reason, the coverage was not run.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/component/AppConfigurationV2StructureProvider.java
```diff
@@ -0,0 +1,64 @@
+
+
```
hilmarf, line 2:
```suggestion
```

services/distribution/src/main/resources/application.yaml
```diff
@@ -16,6 +16,8 @@ services:
   distribution:
     # The name of the distribution output file.
     output-file-name: index
+    # The name of the distribution output file v2.
+    output-file-name-v2: index-v2
```
hilmarf, line 20: `index-v2` ??? where is this used? the path of the configs should be a different one... I don't think we should have any file, that has index-v2 as file name, right?

services/distribution/src/main/resources/main-config/v2/risk-calculation-parameters-1.15.yaml
```diff
@@ -0,0 +1,94 @@
...
+      min: 55
+      max: 63
+      max-exclusive: true
+    weight: 0.5
+  -
+    attenuation-range:
+      min: 63
+      max: 73
+      max-exclusive: true
+    weight: 0.3
```
hilmarf, line 32: @mlenkeit / @ioangut - is the yaml correct? does it work properly in regards to the array?

hilmarf (approved):

services/distribution/src/main/resources/application.yaml
ioangut: We need this file because this is the way the directory structure of the v1, v2 config is developed.

services/distribution/src/main/resources/main-config/v2/risk-calculation-parameters-1.15.yaml
ioangut: I think yes.  Max reviewed it also and add some changes.

PR #1235, opened 2021-03-02
Feature/download from more than two fg

FelixRottler:
- introduces download from more than one gateway
- this pr adds two new profiles "connect-efgs" and "connect-chgs" that run the download in different configurations. Depending on the profile a discriminator column keeps tracks from which target system the federation batch info comes from


common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/FederationBatchTarget.java
```diff
@@ -0,0 +1,8 @@
+package app.coronawarn.server.common.persistence.domain;
+
+public enum FederationBatchTarget {
+  EFGS,
```
EugenM-SAP, line 4: Maybe worth to have the unabbreviated explanations on top of each enum

services/download/src/main/resources/application.yaml
```diff
@@ -14,12 +14,13 @@ logging:
 
 services:
   download:
+    target-system: EFGS
```
EugenM-SAP, line 17: I would think that having default target system is not ideal. I think we should enforce that a specific connect-xyz profile is always specified and fail fasst otherwise. What do you think?

services/download/src/main/resources/bootstrap-cloud.yaml
```diff
@@ -5,7 +5,7 @@ spring:
   cloud:
     vault:
       ssl:
-        trust-store: file:${SSL_VAULT_TRUSTSTORE_PATH}
```
EugenM-SAP, line 8: Are we sure the truststore  is not injected in the container file system?

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/FederationBatchInfo.java
```diff
@@ -16,27 +16,33 @@
...
+   * @param status   status stored as {@link FederationBatchStatus}
    */
   @PersistenceConstructor
-  public FederationBatchInfo(String batchTag, LocalDate date, FederationBatchStatus status) {
+  public FederationBatchInfo(String batchTag, LocalDate date, FederationBatchStatus status,
+      FederationBatchTarget targetSystem) {
     this.batchTag = batchTag;
     this.date = date;
     this.status = status;
+    this.targetSystem = targetSystem;
```
EugenM-SAP, line 45: I know we referred to the gateway as a target system during discussions, but in the case of Download it represents rather a "source" system

services/upload/src/main/resources/bootstrap-cloud.yaml
```diff
@@ -5,7 +5,7 @@ spring:
   cloud:
     vault:
       ssl:
-        trust-store: file:${SSL_VAULT_TRUSTSTORE_PATH}
```
EugenM-SAP, line 8: same as question above

common/persistence/src/main/resources/db/migration/V8__alterFederationBatchInfoAddSource.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE federation_batch_info
+  ADD source_system VARCHAR(10) NOT NULL;
```
hilmarf, line 2: I doubt, that this works. Adding new column with `NOT NULL` but without default!?!

services/download/src/main/resources/application.yaml
FelixRottler: ok

services/download/src/main/resources/bootstrap-cloud.yaml
FelixRottler: I did this because vault ssl was not working on dev when using the prefix "file:". Maybe we need to investigate this

common/persistence/src/main/resources/db/migration/V8__alterFederationBatchInfoAddSource.sql
FelixRottler: True. Will change it

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/FederationBatchSourceSystem.java
```diff
@@ -0,0 +1,13 @@
+package app.coronawarn.server.common.persistence.domain;
+
+/**
+ * Source for downloading  federation batches from different sources.
+ */
+public enum FederationBatchSourceSystem {
+  // EFGS = European Federation Gateway Service
+  EFGS,
+  // CHGS = Swiss Gateway Service
```
hilmarf, line 9: why no real javadoc?

hilmarf (approved):

ioangut (commented):
runconfigs/IntelliJ/callback [spring-boot_run...].run.xml, do we need to add these changes into the callback run config? Please remove this file from this PR if not needed.

services/callback/src/main/java/app/coronawarn/server/services/callback/controller/CallbackController.java
```diff
@@ -44,7 +46,8 @@ public CallbackController(FederationBatchInfoService federationBatchInfoService)
   public ResponseEntity<Void> handleCallback(@RequestParam String batchTag,
       @NotNull @DateTimeFormat(iso = ISO.DATE) @RequestParam LocalDate date) {
     logger.info("BatchInfo with tag {} and date {} received from federation gateway.", batchTag, date);
-    FederationBatchInfo federationBatchInfo = new FederationBatchInfo(batchTag, date);
+    // TODO check how to integrate swiss?
```
ioangut, line 49: Would this TODO be developed in another PR ?

services/download/src/main/java/app/coronawarn/server/services/download/Application.java
```diff
@@ -34,6 +34,8 @@ public static void main(String[] args) {
     SpringApplication.run(Application.class);
   }
 
+
+
```
ioangut, line 38: Please remove these empty lines.

ioangut (approved):

EugenM-SAP (approved):

services/callback/src/main/java/app/coronawarn/server/services/callback/controller/CallbackController.java
FelixRottler: yes @EugenM-SAP

PR #1247, opened 2021-03-10
Event Registration

hilmarf:
let's keep track of the branch...


roesslerj (approved):

pom.xml
```diff
@@ -48,7 +48,7 @@
     <sonar.projectKey>${sonar.organization}_cwa-${project.artifactId}</sonar.projectKey>
 
     <!-- https://spring.io/projects/spring-boot -->
-    <spring-boot.version>2.4.4</spring-boot.version>
+    <spring-boot.version>2.4.3</spring-boot.version>
```
ioangut, line 51: We change the version of spring boot to 2.4.4. Maybe we should keep it.

hilmarf: @ioangut - right. Main branch versions are running fine, correct? Then please keep the newer version

FelixRottler: it should be the same for spring cloud version then @ioangut

ioangut (approved):

pom.xml
ioangut: I checked and Spring Cloud version is the same @FelixRottler

PR #1270, opened 2021-03-22
Filter checkins data based on their transmission risk level. mapping

EugenM-SAP:
## Checklist

## Description

Any check-in with a transmissionRiskLevel that maps to a Transmission Risk Value (as per app-config) of 0 (or 0.0) shall be excluded from further processing (i.e. filtered out). This implies that the risk calculation paramters / trl mapping config added to 1.15 v2 gets moved to a common place that can be shared between Distribution and Submission service.

### DIFF
--- a/common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/config/PreDistributionTrlValueMappingProvider.java
+++ b/common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/config/PreDistributionTrlValueMappingProvider.java
@@ -1,10 +1,10 @@
 package app.coronawarn.server.common.persistence.domain.config;
 
+import app.coronawarn.server.common.persistence.utils.YamlPropertySourceFactory;
 import java.util.List;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.PropertySource;
-import app.coronawarn.server.common.persistence.utils.YamlPropertySourceFactory;
 
 @Configuration
 @ConfigurationProperties()
--- a/common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/config/PreDistributionTrlValueMappingProvider.java
+++ b/common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/config/PreDistributionTrlValueMappingProvider.java
@@ -2,6 +2,7 @@ package app.coronawarn.server.common.persistence.domain.config;
 
 import app.coronawarn.server.common.persistence.utils.YamlPropertySourceFactory;
 import java.util.List;
+import java.util.stream.Collectors;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.PropertySource;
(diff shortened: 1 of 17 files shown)


common/persistence/src/main/resources/trl-value-mapping-1.15.yaml
```diff
@@ -0,0 +1,25 @@
+transmission-risk-value-mapping:
+  -
+    transmission-risk-level: 1
+    transmission-risk-value: 0
+  -
+    transmission-risk-level: 2
+    transmission-risk-value: 0
```
hilmarf, line 7: those will be then the ones, that will be filtered out during `submission`, right?

services/submission/src/main/java/app/coronawarn/server/services/submission/checkins/EventCheckinDataFilter.java
```diff
@@ -61,4 +68,11 @@ private boolean filterOutFutureCheckins(CheckIn checkin) {
   private boolean filterByValidSignature(CheckIn checkin) {
     return traceLocationSignatureVerifier.verify(checkin.getSignedLocation());
   }
+
+  private boolean mapsTo(Integer valueToCheck, Double target) {
+    Optional<TransmissionRiskValueMapping> foundTarget =
+        trlValueMappingProvider.getTransmissionRiskValueMapping().stream()
+            .filter(m -> m.getTransmissionRiskLevel().equals(valueToCheck)).findAny();
+    return foundTarget.isPresent() && foundTarget.get().getTransmissionRiskValue().equals(target);
```
hilmarf, line 76: @EugenM-SAP / @mlenkeit - what about the ones, that are NOT `present`?

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/config/PreDistributionTrlValueMappingProvider.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.stream.Collectors;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.PropertySource;
+
+@Configuration
+@ConfigurationProperties()
+@PropertySource(value = "classpath:trl-value-mapping-1.15.yaml",
+    factory = YamlPropertySourceFactory.class)
+public class PreDistributionTrlValueMappingProvider {
```
hilmarf, line 14: `PreDistributionTrlValueMappingProvider` name is a bit misleading, when we use it during the submission process, right?

common/persistence/src/main/resources/trl-value-mapping-1.15.yaml
EugenM-SAP: @hilmarf  yes

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/config/PreDistributionTrlValueMappingProvider.java
EugenM-SAP: I agree. But somehow that is the context where the mapping is used for. My idea was to put this explicit naming to force us to think if there is a design issue we are missing here. Together with Max we did discuss other aproaches where the configuration can not be shared, but to be honest, the most optimal way is to filter out checkins during submission based on this "distribution-related" configuration. 
We could change the naming or revisit this at a later point in time.

services/submission/src/main/java/app/coronawarn/server/services/submission/checkins/EventCheckinDataFilter.java
EugenM-SAP: In my view, whatever does not meet the TRL mapping value, should be filtered out. Which is what happens above.

hilmarf (approved):

PR #1281, opened 2021-03-29
Protocol Buffer update for release 2.0.x at Mon Mar 29 07:24:43 UTC 2021

Corona-Warn-App-Technical-User:


hilmarf (approved):

services/submission/src/main/java/app/coronawarn/server/services/submission/checkins/EventCheckinDataFilter.java
```diff
@@ -66,7 +66,9 @@ boolean filterOutFutureCheckins(CheckIn checkin) {
   }
 
   boolean filterByValidSignature(CheckIn checkin) {
-    return traceLocationSignatureVerifier.verify(checkin.getSignedLocation());
+    // TODO - @EugenM-SAP do we need still to verify the checkin location?
+    return true;
+    // return traceLocationSignatureVerifier.verify(checkin.getLocationId());
```
hilmarf, line 71: @EugenM-SAP / @FelixRottler - can you please double check if we still need this or how we need to modify, since there is no 'event-reg' anymore

EugenM-SAP: We don't need that part so i will adapt the code.

common/protocols/src/main/proto/app/coronawarn/server/common/protocols/internal/pt/trace_location.proto
```diff
@@ -3,25 +3,36 @@ syntax = "proto3";
...
+  bytes masterPublicKey = 2;
+  bytes cryptographicSeed = 3;
 }
 
 message SignedTraceLocation {
-  bytes location = 1;
-  bytes signature = 2;
+  bytes location = 1 [deprecated = true];
+  bytes signature = 2 [deprecated = true];
 }
```
hilmarf, line 36: @EugenM-SAP - please remove this locally and check which code depends on the same...

ioangut (approved):

PR #1290, opened 2021-03-31
Replace/config file application context initializer

hilmarf:
replace deprecated code with the suggested successors in **tests** only


roesslerj (approved):

services/callback/src/test/java/app/coronawarn/server/services/callback/controller/ApiExceptionHandlerTest.java
```diff
@@ -28,7 +28,7 @@
 
   @BeforeEach
   public void setup() {
-    initMocks(this);
+    openMocks(this);
```
FelixRottler, line 31: We should remove RunWith (JUnit 4) and use instead ExtendWith(MockitoExtension.class,SpringExtension.class) (JUnit 5).
If so then we can:
-  remove the openMocks/initMocks.
- remove the initialization of the callbackController by using InjectMock

FelixRottler (approved):

PR #1298, opened 2021-04-02
Add presence tracing parameters

ioangut:
## Checklist

## Description

Adjust the app config files to include the new provided yaml configuration for Event registration, Presence Trace Parameters.

### DIFF
--- a/services/distribution/src/main/resources/main-config/v2/plausible-deniability-parameters.yaml
+++ b/services/distribution/src/main/resources/main-config/v2/plausible-deniability-parameters.yaml
@@ -1,0 +1,5 @@
+#plausible-deniability-parameters:
+  # validation rules for checkInSizesInBytes:
+  #   * zero entries are allowed
+  #   * each entry must be >= 0
+  check-in-sizes-in-bytes: [10, 20]
--- a/services/distribution/src/main/resources/main-config/v2/presence-tracing-parameters.yaml
+++ b/services/distribution/src/main/resources/main-config/v2/presence-tracing-parameters.yaml
@@ -1,0 +1,121 @@
+#----------------------------------------------------------------------
+# This is the Corona Warn App source configuration file.
+# ----------------------------------------------------------------------
+#
+# This configuration file will be fetched by the mobile app in order to function.
+#
+# Change this file with caution!
+
+risk-calculation-parameters:
+  transmission-risk-value-mapping:
+    -
+      transmission-risk-level: 1
+      transmission-risk-value: 0
+    -
+      transmission-risk-level: 2
+      transmission-risk-value: 0
+    -
+      transmission-risk-level: 3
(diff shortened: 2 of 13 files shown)


services/distribution/src/main/resources/main-config/v2/plausible-deniability-parameters.yaml
```diff
@@ -0,0 +1,5 @@
+#plausible-deniability-parameters:
+  # validation rules for checkInSizesInBytes:
+  #   * zero entries are allowed
+  #   * each entry must be >= 0
+  check-in-sizes-in-bytes: [10, 20]
```
hilmarf, line 5: are we using those values anywhere for check in validation?

services/distribution/src/main/resources/main-config/v2/presence-tracing-parameters.yaml
```diff
@@ -0,0 +1,121 @@
...
+qr-code-descriptors:
+  -
+    regex-pattern: https://e\.coronawarn\.app/c(\d+)/(.+)
+    # validation rule for versionGroupIndex:
+    #   * >= 0
+    version-group-index: 0
+    # validation rule for encodedPayloadGroupIndex:
+    #   * >= 0
+    encoded-payload-group-index: 1
+    payload-encoding: 0 # BASE32
```
hilmarf, line 121: same here... we do have validation in place on submission, which do those checks mentioned in the comments, right?

hilmarf (approved):

hilmarf:
@ioangut don't forget to sync main and release branch!!!

services/distribution/src/main/resources/main-config/v2/plausible-deniability-parameters.yaml
ioangut: Yes, we will use these values to fill up the protobuf for the app config. I used separate file for this one because it is an array of integer and the mapping from yaml to protobuf does not work as expected so I had to deserialise the yaml file.

services/distribution/src/main/resources/main-config/v2/presence-tracing-parameters.yaml
EugenM-SAP: We are using TRL risk value mapping during checkin validation. 

@ioangut  We are already storing those mapping values under persitence/trl-mapping-1.15.yaml. 
We should integrate that yaml (loaded via PreDistributionTrlValueMappingProvider class) into this V2 config when it's being registered as a bean.

ioangut: Even though these values are for the moment similar, they might change differently later. So based on a short discussion with Max, these 2 files should be maintained separately.

roesslerj (approved):

PR #1305, opened 2021-04-06
Fortify issue - close input stream when loading QR template

EugenM-SAP:
## Checklist

* [ ] Make sure `mvn -P integration-tests clean verify` runs for the whole project and, if you touched any code for a service in the `services` folder, ensure it can be run with `spring-boot:run`

## Description

Fixes an issue reported by Fortify scans


services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/qrcode/QrCodeTemplateLoader.java
```diff
@@ -20,8 +21,12 @@ public ByteString loadIosTemplateAsBytes() {
   }
 
   private ByteString loadPosterTemplate(String filename) {
+    InputStream resourceAsStream = null;
     try {
-      return ByteString.readFrom(this.getClass().getClassLoader().getResourceAsStream(filename));
+      resourceAsStream = this.getClass().getClassLoader().getResourceAsStream(filename);
+      ByteString byteString = ByteString.readFrom(resourceAsStream);
+      resourceAsStream.close();
```
roesslerj, line 28: Problem is that the close statement is inside the try-catch block, rendering it useless in case of errors. Probably most elegant fix would be to use a try-with-resources Statement:
`try (InputStream stream =
                   this.getClass().getClassLoader().getResourceAsStream(filename)) {
        return ByteString.readFrom(stream);
}`

roesslerj (dismissed):

hilmarf (approved):

roesslerj (approved):

PR #1306, opened 2021-04-06
Protocol Buffer update for release 2.0.x at Tue Apr  6 08:07:25 UTC 2021

Corona-Warn-App-Technical-User:


EugenM-SAP:
I've added the code adaptions in https://github.com/corona-warn-app/cwa-server/pull/1314
These proto changes have been treated in the same PR above.

PR #1311, opened 2021-04-06
Update QR code poster template parameters

mlenkeit:
This PR updates the QR code poster templates with some small number changes as requested by the dev teams.

@EugenM-SAP the new default values have float values instead of integer. Are these parsed correctly?


roesslerj (approved):

EugenM-SAP:
@mlenkeit  Not parsed in the current implementation but we will adapt it for the new float values

EugenM-SAP (approved):

PR #1326, opened 2021-04-08
Enhance logging when filtering checkins

EugenM-SAP:
## Checklist

* [ ] Make sure `mvn -P integration-tests clean verify` runs for the whole project and, if you touched any code for a service in the `services` folder, ensure it can be run with `spring-boot:run`

### DIFF
--- a/services/submission/src/main/java/app/coronawarn/server/services/submission/checkins/EventCheckinDataFilter.java
+++ b/services/submission/src/main/java/app/coronawarn/server/services/submission/checkins/EventCheckinDataFilter.java
@@ -46,7 +46,7 @@ public class EventCheckinDataFilter {
     AtomicInteger checkinsPickedAfterTrlFiltering = new AtomicInteger();
     AtomicInteger checkinsPickedAfterOldFiltering = new AtomicInteger();
     AtomicInteger checkinsPickedAfterFutureFiltering = new AtomicInteger();
-    var filtered = checkins.stream()
+    final var filtered = checkins.stream()
         .filter(this::filterOutZeroTransmissionRiskLevel)
         .peek(k -> checkinsPickedAfterTrlFiltering.incrementAndGet())
         .filter(this::filterOutOldCheckins)
--- a/services/submission/src/main/java/app/coronawarn/server/services/submission/checkins/EventCheckinDataFilter.java
+++ b/services/submission/src/main/java/app/coronawarn/server/services/submission/checkins/EventCheckinDataFilter.java
@@ -5,17 +5,23 @@ import static java.time.ZoneOffset.UTC;
 
 import app.coronawarn.server.common.persistence.domain.config.PreDistributionTrlValueMappingProvider;
 import app.coronawarn.server.common.persistence.domain.config.TransmissionRiskValueMapping;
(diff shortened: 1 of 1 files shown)


ioangut (dismissed):

services/submission/src/main/java/app/coronawarn/server/services/submission/checkins/EventCheckinDataFilter.java
```diff
@@ -37,10 +43,24 @@ public EventCheckinDataFilter(SubmissionServiceConfig submissionServiceConfig,
...
+    AtomicInteger checkinsPickedAfterOldFiltering = new AtomicInteger();
+    AtomicInteger checkinsPickedAfterFutureFiltering = new AtomicInteger();
+    var filtered = checkins.stream()
         .filter(this::filterOutZeroTransmissionRiskLevel)
+        .peek(k -> checkinsPickedAfterTrlFiltering.incrementAndGet())
         .filter(this::filterOutOldCheckins)
-        .filter(this::filterOutFutureCheckins).collect(Collectors.toList());
+        .peek(k -> checkinsPickedAfterOldFiltering.incrementAndGet())
+        .filter(this::filterOutFutureCheckins)
+        .peek(k -> checkinsPickedAfterFutureFiltering.incrementAndGet())
```
pithumke, line 55: Peek is not intended for code with side-effects and this will most likely break in Java 9 (see [here](https://stackoverflow.com/a/48221925)). Although it doesn't look as nice as your full-on stream solution, you should probably consider something like this:

```Java
int checkinsBeforeFiltering = checkins.size();

var List<CheckIn> filtered = checkins.stream().filter(this::filterOutZeroTransmissionRiskLevel).collect(Collectors.toList());
int checkinsPickedAfterTrlFiltering = filtered.size();

filtered = checkins.stream().filter(this::filterOutOldCheckins).collect(Collectors.toList());
int checkinsPickedAfterOldFiltering = filtered.size();

filtered = checkins.stream().filter(this::filterOutFutureCheckins).collect(Collectors.toList());
int checkinsPickedAfterFutureFiltering = filtered.size();
```

pithumke (changes requested):

services/submission/src/main/java/app/coronawarn/server/services/submission/checkins/EventCheckinDataFilter.java
pithumke: On the other hand, the same is already being done in the [FederationUploadKeyService](https://github.com/corona-warn-app/cwa-server/blob/b75ba8e12bff38c878f4ada328c1cf8b08559281/common/persistence/src/main/java/app/coronawarn/server/common/persistence/service/FederationUploadKeyService.java#L60). So, let's not start a refactoring now.

pithumke (approved):

services/submission/src/main/java/app/coronawarn/server/services/submission/checkins/EventCheckinDataFilter.java
EugenM-SAP: Good point though. Thanks for raising awareness. It kind of falls in the technical debt backlog for now.

PR #1349, opened 2021-04-20
Load the QRCode template dynamically via env var

roesslerj:
This loades the QR Code template dynamically, using an env var. If it's not set, it will still load the one from the repo...

https://github.com/corona-warn-app/cwa-admin/pull/164
EXPOSUREBACK-1079


hilmarf (approved):
What about necessary helm-chart or vault changes?

roesslerj:
Helm char values are adapted here: https://github.com/corona-warn-app/cwa-admin/pull/164

ioangut (approved):

PR #1354, opened 2021-04-23
Refactor/shared module

alexmoroianu:


common/shared/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+  </parent>
+
+  <artifactId>shared</artifactId>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.opencwa</groupId>
+      <artifactId>protocols</artifactId>
+    </dependency>
+    <dependency>
```
FelixRottler, line 19: I think a lot of these dependencies are not needed inside this project. Maybe we can clean this pom.xml up a little bit

common/shared/src/main/java/app/coronawarn/server/common/shared/exception/UnableToLoadFileException.java
```diff
@@ -0,0 +1,17 @@
+
+
+package app.coronawarn.server.common.shared.exception;
+
+/**
+ * The file could not be loaded/parsed correctly.
+ */
+public class UnableToLoadFileException extends Exception {
```
FelixRottler, line 8: In the distribution service this class exists too ( and is actually used)

common/shared/src/main/java/app/coronawarn/server/common/shared/util/HashUtils.java
```diff
@@ -0,0 +1,66 @@
...
+   * that implements the chosen {@link MessageDigestAlgorithms}.
+   *
+   * @param data - ByteString to be diggested
+   * @return - digested ByteString
+   */
+  public static byte[] byteStringDigest(ByteString data, MessageDigestAlgorithms algorithm) {
+    try {
+      return MessageDigest.getInstance(algorithm.getName()).digest(data.toByteArray());
+    } catch (NoSuchAlgorithmException e) {
+      // DO NOTHING
```
FelixRottler, line 52: Empty catch

common/shared/src/test/java/app/coronawarn/server/common/shared/util/HashUtilsTest.java
```diff
@@ -0,0 +1,32 @@
...
+import org.junit.jupiter.api.Test;
+import java.security.NoSuchAlgorithmException;
+
+class HashUtilsTest {
+
+  public static final String TEST_STRING = "string";
+
+  @Test
+  void testMd5DigestAsHex() {
+    String hash = md5DigestAsHex(TEST_STRING);
```
FelixRottler, line 17: Maybe we can make this test more useful. We could take an existing hex md5 hash (and its origin) and see if this function gives the same result.

common/shared/src/test/java/app/coronawarn/server/common/shared/util/HashUtilsTest.java
```diff
@@ -0,0 +1,32 @@
...
+    assertThat(hash).hasSize(32);
+  }
+
+  @Test
+  void testRandomByteArrayData() {
+    byte[] hash = generateSecureRandomByteArrayData(16);
+    assertThat(hash).hasSize(16);
```
FelixRottler, line 24: Maybe we can also refactor this test. The advantage of the secure random is that even if we use the same seed for two different random variables the result will be different

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/structure/directory/IndexDirectory.java
```diff
@@ -2,9 +2,9 @@
 
 package app.coronawarn.server.services.distribution.assembly.structure.directory;
 
+import app.coronawarn.server.common.shared.collection.ImmutableStack;
+import app.coronawarn.server.common.shared.functional.Formatter;
 import app.coronawarn.server.services.distribution.assembly.structure.Writable;
-import app.coronawarn.server.services.distribution.assembly.structure.util.ImmutableStack;
-import app.coronawarn.server.services.distribution.assembly.structure.util.functional.Formatter;
 import app.coronawarn.server.services.distribution.assembly.structure.util.functional.WritableFunction;
```
FelixRottler, line 8: Is there a reason that WritableFunction is not extracted into the shared module?

services/submission/src/test/java/app/coronawarn/server/services/submission/checkins/EventCheckinsFacadeIT.java
```diff
@@ -52,17 +41,6 @@ public void setup() {
     currentTimestamp = CheckinsDateSpecification.HOUR_SINCE_EPOCH_DERIVATION.apply(Instant.now().getEpochSecond());
   }
 
-  @Test
```
FelixRottler, line 55: Why is that Test removed?

FelixRottler (changes requested):
see comments :)

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/structure/directory/IndexDirectory.java
alexmoroianu: Hi Felix. Actually, the entire "structure" package would make sense to be moved into an "io" package into shared module but I would do this in the next iteration as there are dozens of files and would really complicate this PR. What do you think ?

services/submission/src/test/java/app/coronawarn/server/services/submission/checkins/EventCheckinsFacadeIT.java
alexmoroianu: I reverted it back.

common/shared/pom.xml
alexmoroianu: Removed them.

common/shared/src/main/java/app/coronawarn/server/common/shared/util/HashUtils.java
alexmoroianu: Added a logger.

common/shared/src/main/java/app/coronawarn/server/common/shared/exception/UnableToLoadFileException.java
alexmoroianu: Yes, indeed. I removed it from distribution and referenced it from shared.

common/shared/src/test/java/app/coronawarn/server/common/shared/util/HashUtilsTest.java
alexmoroianu: Done.

ioangut (approved):

FelixRottler (approved):

PR #1362, opened 2021-04-28
Hf/2.1.3

hilmarf:
EXPOSUREBACK-1209


ioangut (dismissed):

mlenkeit (dismissed):
LGTM, anyway needs to be confirmed on pre-PROD.

floscher:
@hilmarf What is the status of this pull request? Commit https://github.com/corona-warn-app/cwa-server/commit/a7b80cb2d1b68231e7250d54408f3ba486751ff9 is needed for https://github.com/corona-warn-app/cwa-documentation/issues/590 , so the newer QR codes of the luca app can be read by CWA.

Ein-Tim:
Hey @thomasaugsten 

Could you maybe elaborate a bit where you see the potential that a CWA user gets deanonymized? 
This would be nice, thank you!

thomasaugsten:
If the 3rd Party saves the CWA Event Codes he can connect the real address to an published Event codes and deanonymize the published event. If a low number of people checks in the event he can easily figure out which person was tested positive because he can access the real contact data of the person.

OlympianRevolution:
That may be the case when warning the event through the CWA. But when the event warns the CWA user there is no deanonymization right? Could the direction event to user be activated at least? And just give a fat warning in the direction user to event when uploading to a non CWA event that someone with access to nonanonymous lists may know it was you.

(Isn't you argument true for entire counties now? Where only very few people test positive, someone with full access to Gesundheitsamt lists can guess where a warning came from. This has always been the case for people that generally have few contacts. They can always guess who it was. What is different here? I still don't see it.

Bad actor has access to lists and event gets warned. They can go through lists and try to use other info (such as?) to guess who it was. Does a user have the expectation to be anonymous when they entered their name on a list?

Is this any different from a CWA event where someone also keeps a paper list because they are required to?

Sorry for the bad form. But many users are complaining about this because it was partially communicated it would work.

neugens:
> If the 3rd Party saves the CWA Event Codes he can connect the real address to an published Event codes and deanonymize the published event. If a low number of people checks in the event he can easily figure out which person was tested positive because he can access the real contact data of the person.

This is true, however the current situation is that LucaApp is being forced to everyone, very few are using the CoronaWarn QR unfortunately, and even less shops are willing to listen to this feedback, the alternative is either sign by hand or worse without any chance to ever get proper GDPR treatment for our data. Having some form of compatibility would also help increase the number of people using the CoronaWarn app. We can argue why the State has decided to give priority to an insecure and proprietary app that hijack the public one, but essentially that's where we are and we really need a solution!

Ein-Tim:
I think it's really important to differentiate here. I understand everything who says that the compatibility with luca should just be published now. I'm sure there would be many reports and happy users who would be very thankful. But on the other hand, we saw how the luca app was criticized from all sides and trust was lost. Imagine reports from the CCC which "reveal" that it is possible to deanonymize CWA users.

But moreover, the question is how a implementation of the CWA with luca would work. 

Would the CWA users check in to an luca QR code and only get warned if other checked-in CWA users warn their contacts?
Would the CWA users check in to an luca QR code and also get warned if a person who used luca to check in warns their contacts? (I don't use luca - did they implement the warning via app function in the meantime or does the health authority still only get a list of check in people and they call them?)
Would the CWA users check in anonym? Would this be in compliance with the current "Verordnungen"? 

AFAIK The luca app contacts a server every time someone checks in (thus no offline check in) - would the CWA also has to contact to this server? If not, how should cross system warnings work?

I think it's more than just making the QR-Code scannable for CWA users, there are many questions which have to be clarified.

Ein-Tim:
The only question I have (cc @thomasaugsten) is whether the implementation is still planned or if work on this feature has stopped. I the latter case I suggest to close this PR.

thomasaugsten:
The feature has a good progress. We hope we can finish the feature soon.

BITS-Editor:
Good news! Thank you!

vaubaehn:
According to a post of the Android Store support team, the technical works have been finished, everything is prepared - it's not only up to a review/decision of 'authorities'...
Is it with the Datenschützer or with BSI currently?

rugk:
Also, BTW, can anyone clarify that strange PR title? `Hf/2.1.3` could not be a less obscure code for "Add luca app to whitelisted QR codes for even registration."… :monocle_face:

thomasaugsten:
Hotfix version 2.1.3

rugk:
Oh, then this "hotfix" will soon celebrate it's 6month birthday… :upside_down_face:

ndegendogo:
> Hotfix version 2.1.3

Ah - and I thought it is the initials of the author's name 😅😅

xandro0777:
> If the 3rd Party saves the CWA Event Codes he can connect the real address to an published Event codes and deanonymize the published event. If a low number of people checks in the event he can easily figure out which person was tested positive because he can access the real contact data of the person.

There are more ways to attack the anonymization provided by CWA: 
* an evil employer can place an Android device with CWA in every room getting near perfect de-anonymization trivially. 
* any user can pin down the near exact time of the contact with an infected by using several devices with CWA switched on/of at various times. With just 10 phones you can distinguish 1024 discrete events or time slots. 

However in my opinion forcing users to use Luca app is the biggest threat : here the threat is not only de-anonymization in case of an infection event but also for example Luca server infrastructure could be hacked leaking private data and movement details of all Luca app users.

Anything that can be done to end the monopoly of Luca app would be a great improvement even if perfect anonymity can't be granted.

vaubaehn:
For 1 second I actually thought, there is a positive progress... But no...

ioangut:
We will close this PR for the moment and create a new one this week for the actual release we are developing right now.

Ein-Tim:
New PR: #1594

DanAmt:
IMHO the compatibility of these codes can be important to digital contact tracing  and therefore pandemic control.
Media is starting to ask about this, so it would be good to have a defined state and a poss release date.

PR #1363, opened 2021-04-29
please avoid empty catch blocks...

hilmarf:
even if the situation should theoretically never happen ;-)


alexmoroianu (approved):
Okay, got it.

FelixRottler (approved):

PR #1393, opened 2021-05-19
Protocol Buffer update for release 2.4.x at Wed May 19 06:57:49 UTC 2021

Corona-Warn-App-Technical-User:


roesslerj (approved):
I don't think this is a big change, but I'm still fine with it ;-)

hilmarf:
empty change

PR #1416, opened 2021-06-03
adjust primary key

hilmarf:
when we store the batches of two different systems (EFGS/CHGS), we
should adjust our primary key!


alexmoroianu (approved):

hilmarf:
#1414

PR #1417, opened 2021-06-03
FEATURE - Local statistics

alexmoroianu:


roesslerj (approved):

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/LocalStatisticsDownload.java
```diff
@@ -0,0 +1,37 @@
+package app.coronawarn.server.common.persistence.domain;
+
+import org.springframework.data.annotation.Id;
+
+public class LocalStatisticsDownload {
```
FelixRottler, line 5: Mismatch between table name and entity. So the repository non-custom methods like findAll() will fail because they look for the wrong table.

common/persistence/src/main/java/app/coronawarn/server/common/persistence/service/LocalStatisticsDownloadService.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  /**
+   * Retrieves the latest {@link LocalStatisticsDownload} stored. The order is determined by the counter field, which is
+   * automatically incremented by the Database.
+   *
+   * @return {@link LocalStatisticsDownload} returns Optional.empty if no download entries are stored.
+   */
+  public Optional<LocalStatisticsDownload> getMostRecentDownload() {
+    return Optional.ofNullable(repository.getWithLatestETag());
```
FelixRottler, line 51: What do you think of moving the Optional into the Repository? So basicall y in the repository return   Optional<LocalStatisticsDownload>

common/persistence/src/main/java/app/coronawarn/server/common/persistence/service/LocalStatisticsDownloadService.java
```diff
@@ -0,0 +1,72 @@
...
+   * value automatically determined by a postgres sequence.
+   *
+   * @param timestamp in seconds.
+   * @param etag      value to be stored.
+   * @return boolean determining if store was successful or not.s
+   */
+  @Transactional
+  public boolean store(long timestamp, String etag) {
+    try {
+      this.repository.insertWithAutoIncrement(timestamp, etag);
```
FelixRottler, line 36: I think also the insert statement of the repository can return an boolean which indicates if the insertion was successful

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/component/LocalStatisticsStructureProvider.java
```diff
@@ -0,0 +1,52 @@
...
+      Map<Integer, LocalStatistics> localStatisticsMap) {
+    this.cryptoProvider = cryptoProvider;
+    this.distributionServiceConfig = distributionServiceConfig;
+    this.localStatisticsMap = localStatisticsMap;
+  }
+
+  /**
+   * Returns a list containing the archives with Local Statistics for mobile clients using signature file.
+   */
+  public List<Writable<WritableOnDisk>> getLocalStatisticsList() {
```
FelixRottler, line 40: Not sure about the structure here. Do we not normally return a Directory<WriteableOnDisk> when working with multiple data? For example  public Directory<WritableOnDisk> getDiagnosisKeys() {. What is the difference here?

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/local/BuildLocalStatisticsHelper.java
```diff
@@ -0,0 +1,238 @@
...
+   * Find federal state code from province code.
+   * For province codes containing 4 digits, the federal state code consists of first digit.
+   * For province codes containing 5 digits, the federal state code consists of first two digits.
+   * For province codes containing 1 or 2 digits, the federal state code is the the same as province code.
+   * @param provinceCode - province code.
+   * @return - federal state code.
+   */
+  public static Optional<String> findFederalStateByProvinceCode(String provinceCode) {
+    switch (provinceCode.length()) {
+      case 1:
```
FelixRottler, line 116: Is there a return statement missing?

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/local/LocalStatisticsToProtobufMapping.java
```diff
@@ -0,0 +1,233 @@
...
+    Map<Integer, LocalStatistics> localStatisticsMap = new HashMap<>();
+
+    try {
+      Optional<JsonFile> file = this.getFile();
+
+      if (file.isEmpty()) {
+        logger.warn("Stats file is already updated to the latest version. Skipping generation.");
+        return Collections.emptyMap();
+      } else {
+        this.updateETag(file.get().getETag());
```
FelixRottler, line 79: Can we unwrap the optional here? and use the file itself for function parameters? For example the deserializeAndValidate

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/local/LocalStatisticsToProtobufMapping.java
```diff
@@ -0,0 +1,233 @@
...
+   *
+   * @param localStatisticsMap - local statistics map grouped by archive index.
+   * @param federalStateCode - federal state code
+   * @param localStatisticsJsonStringObject - local statistics json object
+   * @param statisticsSupplier - supplier which builds local statistics
+   * @param statisticsEnhancer - supplier which enhance local statistics
+   */
+  private void fillLocalStatisticsFederalStatesGroupMap(Map<Integer, LocalStatistics> localStatisticsMap,
+      int federalStateCode,
+      LocalStatisticsJsonStringObject localStatisticsJsonStringObject,
```
FelixRottler, line 157: I think this parameter is not used or?

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/local/LocalStatisticsToProtobufMapping.java
```diff
@@ -0,0 +1,233 @@
...
+    }
+  }
+
+  private void updateETag(String newETag) {
+    var currentTimestamp = TimeUtils.getCurrentUtcHour().toEpochSecond(ZoneOffset.UTC);
+    this.localStatisticsDownloadService.store(currentTimestamp, newETag);
+  }
+
+  /**
+   * Converts an {@link LocalStatisticsJsonStringObject} into a Local Statistics object and handle it addition
```
FelixRottler, line 135: Since the parameter is not used, i do not understand the comment. Maybe you could explain it to me :)

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/local/LocalStatisticsToProtobufMapping.java
```diff
@@ -0,0 +1,233 @@
...
+   * @return map containing local statistics grouped by archive id.
+   */
+  @Bean
+  public Map<Integer, LocalStatistics> constructProtobufLocalStatistics() {
+    Map<Integer, LocalStatistics> localStatisticsMap = new HashMap<>();
+
+    try {
+      Optional<JsonFile> file = this.getFile();
+
+      if (file.isEmpty()) {
```
FelixRottler, line 75: I think there might be an ifPresentOrElse that takes in a consumer and a runnable if the entry does not exist. This would also solve the issue with passing the optional as parameter. What do you think?

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/local/LocalStatisticsToProtobufMapping.java
```diff
@@ -0,0 +1,233 @@
...
+    this.localStatisticsDownloadService = localStatisticsDownloadService;
+    this.regionMappingConfig = regionMappingConfig;
+  }
+
+  /**
+   * Create protobuf local statistic map from raw JSON local statistics.
+   * Local statistics has to be uploaded on CDN in 7 packages. Each entry on the map represents 1 package.
+   * Packages of local statistics contains data about 1 up to 4 federal states and all their administrative units.
+   * For more info related to the grouping of the federal states into packages please look into
+   * https://github.com/corona-warn-app/cwa-app-tech-spec/blob/42e9e4f3c588cd2fd283f904e9f0ccd53a2b83d0/
```
FelixRottler, line 63: Link is dead for me 404

services/distribution/src/main/resources/main-config/region-mapping.yaml
```diff
@@ -0,0 +1,25 @@
...
+# docs/spec/statistics.md#populate-local-statistical-data
+region-mapping:
+  federal-states-groups:
+    0: 4
+    1: 4
+    2: 4
+    3: 4
+    4: 5
+    5: 7
+    6: 7
```
FelixRottler, line 16: 6 should map to one or not?

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/local/BuildLocalStatisticsHelper.java
```diff
@@ -0,0 +1,238 @@
...
+   * Add administrative unit statistics to an already existing Local Statistics from an instance of
+   * {@link LocalStatisticsJsonStringObject}.
+   *
+   * @param localStatistics - existing local statistics
+   * @param federalStateCode - federal state code.
+   * @param localStatisticsJsonStringObject - local statistics json object.
+   * @return - Local Statistics protobuf containing new administrative unit data.
+   */
+  private static LocalStatistics addAdministrativeUnitData(LocalStatistics localStatistics,
+      int federalStateCode,
```
FelixRottler, line 185: unused parameter

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/LocalStatisticsDownload.java
alexmoroianu: Hi Felix. The spring repository is generic, thus the entity class is passed to the interface. It is still a good ideea to have the same name but I followed the statistics_dowloaded table name and entity for consistence purposes.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/component/LocalStatisticsStructureProvider.java
alexmoroianu: Yes, we normally do that but the requirement is for all the archives (ex local_stats_1, local_stats_2, local stats_3 etc) to be in the 'v1' directory and not in their own.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/local/LocalStatisticsToProtobufMapping.java
alexmoroianu: Checking with a normal if gives me the possibility to return an empty map or the filled map. I don't know if I can return the bean value from a functional method. But I get rid of the optional because I can call optional.get() in the else condition. What do you think ?

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/local/LocalStatisticsToProtobufMapping.java
alexmoroianu: Yes, that's what I mentioned above :D I do that.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/local/LocalStatisticsToProtobufMapping.java
alexmoroianu: What a good remark. I adapted the javadoc and removed the parameter.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/local/LocalStatisticsToProtobufMapping.java
alexmoroianu: Indeed. Removed

services/distribution/src/main/resources/main-config/region-mapping.yaml
alexmoroianu: No, that's a mistake in the specs. If you see there are two states with id 7 BW and RP which is wrong. BW is actually fedderal state 8

alexmoroianu: But anyway, remarkable PR skills 🥇

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/local/BuildLocalStatisticsHelper.java
alexmoroianu: No, not really. Case 1 and 2 have the same return.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/local/BuildLocalStatisticsHelper.java
alexmoroianu: Indeed. Removed.

common/persistence/src/main/java/app/coronawarn/server/common/persistence/service/LocalStatisticsDownloadService.java
alexmoroianu: I agree. I moved it.

common/persistence/src/main/java/app/coronawarn/server/common/persistence/service/LocalStatisticsDownloadService.java
alexmoroianu: Agree. Refactored it.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/file/MockStatisticJsonFileLoader.java
```diff
@@ -10,7 +11,7 @@
 
 @Component
 @Profile("local-json-stats")
-public class LocalStatisticJsonFileLoader implements JsonFileLoader {
+public class MockStatisticJsonFileLoader implements StatisticJsonFileLoader {
```
hilmarf, line 14: what is it 'mocking'? why we do have 'mocks' in productive code?

common/persistence/src/main/java/app/coronawarn/server/common/persistence/domain/LocalStatisticsDownload.java
FelixRottler: I dont understand? If you execute findAll() with the generic repository you created for local statistics then it will fail because it will lookup the wrong table

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/local/LocalStatisticsToProtobufMapping.java
FelixRottler: agree

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/file/MockStatisticJsonFileLoader.java
alexmoroianu: Hi Hilmar. This is the statistics json file loader for local machine. It was already there when I started implementing. I renamed it from LocalStatistics to MockStatistics in order to avoid confusion because now we have all this classes starting with LocalStatistics which in fact mean "Regional" or "Federal state grouped". This class used the LocalStatistic naming to refer to "Local" environment.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/component/LocalStatisticsStructureProvider.java
```diff
@@ -0,0 +1,52 @@
...
+    if (!localStatisticsMap.isEmpty()) {
+      return localStatisticsMap
+          .keySet()
+          .stream()
+          .map(key -> new LocalStatisticsDirectory(cryptoProvider, distributionServiceConfig,
+              localStatisticsMap.get(key)).getLocalStatisticsArchive(key))
+          .collect(Collectors.toList());
+    }
+
+    return null;
```
hilmarf, line 50: wouldn't this result in NullPointerException? Is it intended? Top priority has the key-file distribution!!! This must NOT under any circumstances fail... so in other words, if the stats are not working, it's bad, but not serious - NPEs should NOT stop the TEKs distribution

```suggestion
    return Collections.emptyList();
```

alexmoroianu: yes, indeed it could lead to a NullPointer. I have updated the code and followed your recommendation.l

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/local/LocalStatisticsToProtobufMapping.java
```diff
@@ -0,0 +1,228 @@
...
+
+    try {
+      Optional<JsonFile> optionalFile = this.getFile();
+
+      if (optionalFile.isEmpty()) {
+        logger.warn("Stats file is already updated to the latest version. Skipping generation.");
+        return Collections.emptyMap();
+      } else {
+        JsonFile file = optionalFile.get();
+        this.updateETag(optionalFile.get().getETag());
```
hilmarf, line 80: the `updateETag` is our save point, right? Shouldn't this then moved to the end of the processing and only happen, if nothing else fails before?

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/local/LocalStatisticsToProtobufMapping.java
```diff
@@ -0,0 +1,228 @@
...
+
+        List<LocalStatisticsJsonStringObject> onePerProvinceStatistics = deserializeAndValidate(file);
+
+        onePerProvinceStatistics.forEach(localStatisticsJsonStringObject -> {
+          if (localStatisticsJsonStringObject.getProvinceCode() != null) {
+            int provinceCode = Integer.parseInt(localStatisticsJsonStringObject.getProvinceCode());
```
hilmarf, line 86: 👍

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/local/LocalStatisticsToProtobufMapping.java
```diff
@@ -0,0 +1,228 @@
...
+
+            if (isFederalState(localStatisticsJsonStringObject.getProvinceCode())) {
```
hilmarf, line 87: if you have parsed already the province code into `int`, why then not operating on the same?

`isFederalState(int i) { return i < 100; } `

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/local/LocalStatisticsToProtobufMapping.java
alexmoroianu: Yes, I agree. I moved it at the end.

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/local/LocalStatisticsToProtobufMapping.java
```diff
@@ -0,0 +1,228 @@
...
+              fillLocalStatisticsFederalStatesGroupMap(
+                  localStatisticsMap,
+                  provinceCode,
+                  federalStateSupplier(provinceCode, localStatisticsJsonStringObject),
+                  federalStateEnhancer(provinceCode, localStatisticsJsonStringObject)
+              );
+            } else {
+
+              Optional<String> federalStateCodeOptional = findFederalStateByProvinceCode(
```
hilmarf, line 96: `findFederalStateByProvinceCode(int i) { return i % 1000; } `

alexmoroianu: hmm I agree with this approach, it is more elegant but I think there should be division operation not modulo, right ?
15334 / 1000 = 15 ; 8345 / 1000 = 8.
15334 % 1000 = 334 ; 8345 / 1000 = 345

services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/local/LocalStatisticsToProtobufMapping.java
alexmoroianu: Yes, I will change it to this.

hilmarf (approved):

PR #1456, opened 2021-06-24
add bootstrap vault configuration for different profiles

ioangut:
## Checklist

## Description

Adapt bootstrap configuration for swiss federation integration.

### DIFF
--- a/services/callback/src/main/resources/bootstrap-cloud.yaml
+++ b/services/callback/src/main/resources/bootstrap-cloud.yaml
@@ -14,7 +14,6 @@ spring:
         enabled: true
         backend: ${VAULT_BACKEND}
         profile-separator: '/'
-        application-name: 'callback'
         default-context: ''
         profiles: cloud
       fail-fast: true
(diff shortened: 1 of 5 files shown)


services/callback/src/main/resources/bootstrap-cloud.yaml
```diff
@@ -14,7 +14,6 @@ spring:
         enabled: true
         backend: ${VAULT_BACKEND}
         profile-separator: '/'
-        application-name: 'callback'
         default-context: ''
         profiles: cloud
```
hilmarf, line 18: would this profile then be needed or conflict with the other settings?

hilmarf (approved):

services/callback/src/main/resources/bootstrap-cloud.yaml
FelixRottler: Normally we have in the deploymentConfig something like ACTIVE_PROFILES: cloud, connect-efgs. Removing this 'callback' decouples the vault secret engine from the cloud profile. However it is still the same vault. Therefore I think that the profile here needs to stay.

ioangut: I think is needed and would not conflict with the other settings. This is actually what we are trying to verify, if we have the application name and profiles used  in the federation specific bootstrap yaml the vault error still persists.

FelixRottler (approved):

PR #1471, opened 2021-07-02
Feature/certificate rules

alexmoroianu:


services/distribution/src/main/java/app/coronawarn/server/services/distribution/dgc/DigitalGreenCertificateToCborMapping.java
```diff
@@ -1,49 +1,55 @@
...
+  public DigitalGreenCertificateToCborMapping(DigitalCovidCertificateClient digitalCovidCertificateClient,
+      ResourceLoader resourceLoader) {
     this.digitalCovidCertificateClient = digitalCovidCertificateClient;
+    this.resourceLoader = resourceLoader;
   }
 
   /**
-   * TODO: write javadoc.
+   * Construct country rules retrieved from DCC client for CBOR encoding.
+   *git s
```
ioangut, line 33: Is this a typo?

alexmoroianu: yep :D fixed

ioangut (approved):

AliceBilc (approved):

services/distribution/src/main/java/app/coronawarn/server/services/distribution/dgc/DigitalGreenCertificateToProtobufMapping.java
```diff
@@ -173,12 +174,15 @@ public ValueSets constructProtobufMapping() throws UnableToLoadFileException {
...
+
+    if (result.isEmpty()) {
       String path = configGetter.getPath(distributionServiceConfig.getDigitalGreenCertificate());
       return readConfiguredJsonOrDefault(resourceLoader, path, valueSetDefaultPath,
           ValueSet.class);
-    });
-    return result.orElseThrow(() -> new UnableToLoadFileException(valueSetDefaultPath));
+    }
+
+    return result.get();
```
roesslerj, line 185: Why not throwing a custom exception here (UnableToLoadFileException or DigitalCovidCertificateException)? We don't want the NoSuchElementException here!

services/distribution/src/main/java/app/coronawarn/server/services/distribution/dgc/DigitalGreenCertificateToProtobufMapping.java
```diff
@@ -188,10 +192,16 @@ private ValueSet read(String valueSetId, ConfigValueProvider configGetter, Strin
...
-    if (hash.isPresent()) {
-      return dccClient.getValueSet(hash.get());
+    Optional<String> hash;
+    try {
+      hash = getValueSetHash(valueSetId);
+      if (hash.isPresent()) {
+        return dccClient.getValueSet(hash.get());
+      }
+    } catch (DigitalCovidCertificateException e) {
+      return Optional.empty();
```
roesslerj, line 202: Why not simply declare the thrown exception? Would make all else easier, or not?

roesslerj (changes requested):
I think in general it doesn't make sense to mix two different edge-case handling strategies. Either we should use Optional throughout, or throw exceptions if we can't connect to the servers or can't find the files. Mixing both makes the code complex and error-prone IMHO. What do you think?

roesslerj (approved):

PR #1493, opened 2021-07-15
Add expiration threshold environment variable for app config ios and …

ioangut:
## Checklist

## Description

Add expiration threshold in days environment variable for app config ios and android.


services/distribution/src/main/java/app/coronawarn/server/services/distribution/config/DistributionServiceConfig.java
```diff
@@ -1603,6 +1603,8 @@ public void setRequireEvaluationTypeHardwareBacked(Boolean requireEvaluationType
 
       private DgcTestCertificateParameters dgcTestCertificateParameters;
 
+      private Integer expirationThresholdInDays;
```
hilmarf, line 1606: @mlenkeit - should we define some `@Min(0)` or `@Max(100)` values?

hilmarf (approved):

services/distribution/src/main/java/app/coronawarn/server/services/distribution/config/DistributionServiceConfig.java
mlenkeit: @hilmarf  Good idea!

hilmarf: which values make most sense?

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/appconfig/ApplicationConfigurationV2PublicationConfig.java
```diff
@@ -534,6 +534,8 @@ private DGCParameters buildDgcParameters(
         .build();
     return DGCParameters.newBuilder()
         .setTestCertificateParameters(testCertificateParameters)
+        .setExpirationThresholdInDays(distributionServiceConfig.getAppConfigParameters()
+            .getDgcParameters().getExpirationThresholdInDays())
```
FelixRottler, line 538: Is it possible that ExpirationThresholdInDays will be null? I remember that for some other properties we have a null check ( at least for local stats)

FelixRottler (approved):

PR #1498, opened 2021-07-19
Fix/invalid certificate

alexmoroianu:


services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/component/DigitalGreenCertificateStructureProvider.java
```diff
@@ -69,15 +70,19 @@ private DirectoryOnDisk constructArchiveToPublish(DigitalGreenCertificate dgcCon
...
-      ArchiveOnDisk archiveToPublish = new ArchiveOnDisk(dgcConfig.getValuesetsFileName());
-      archiveToPublish.addWritable(new FileOnDisk(EXPORT_BIN,
-          dgcToProtobufMapping.constructProtobufMapping().toByteArray()));
-      DirectoryOnDisk languageDirectory = new DirectoryOnDisk(currentLanguage.toLowerCase());
-      languageDirectory.addWritable(new DistributionArchiveSigningDecorator(
-          archiveToPublish, cryptoProvider, distributionServiceConfig));
-      dgcDirectory.addWritable(languageDirectory);
-      logger.info("Writing digital green certificate to {}/{}/{}.", dgcDirectory.getName(), languageDirectory.getName(),
-          archiveToPublish.getName());
+      try {
```
FelixRottler, line 73: Is the try/catch inside the for-loop on purpose?

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/component/DigitalGreenCertificateStructureProvider.java
```diff
@@ -98,13 +103,16 @@ private DirectoryOnDisk constructArchiveToPublish(DigitalGreenCertificate dgcCon
       onboardedCountries
           .addWritable(new FileOnDisk("export.bin", dgcToCborMapping.constructCborCountries()));
       logger.info("Onboarded countries archive has been added to the DGC distribution folder");
+
+      return Optional.of(new DistributionArchiveSigningDecorator(onboardedCountries, cryptoProvider,
```
FelixRottler, line 107: I see that `getOnboardedCountriesArchive().ifPresent(dgcDirectory::addWritable);` looks nice, but is there actually a benefit of wrapping this `Optional.of(new DistributionArchiveSigningDecorator(onboardedCountries, cryptoProvider,` in an Optional? The point of failure is `dgcToCborMapping.constructCborCountries()`, so I dont see the benefit here (codingwise) and I would maybe use a Consumer....Maybe we can discuss this :)

services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/component/DigitalGreenCertificateStructureProvider.java
hilmarf: with this, at least there is a chance that some languages don't fail :-)

hilmarf (approved):

AliceBilc (approved):

PR #1502, opened 2021-07-20
adopt default card sequence

hilmarf:
use https://github.com/corona-warn-app/cwa-admin/pull/196 as default


Ein-Tim:
Hi Hilmar, this will actual affect the row in which the statistic cards in the app are shown, am I right?

hilmarf:
> Hi Hilmar, this will actual affect the row in which the statistic cards in the app are shown, am I right?

Hi @Ein-Tim, yes correct. We were asked to put the national 7-day incidence as the first tile in the statistics carousel, which is displayed initially. Actually the `STATS_CARD_ID_SEQUENCE` can be changed as environment variable, which will happen with the deployment of 2.6. This current PR is just to keep our defaults in sync with those environment settings.

Best regards,
Hilmar

ioangut (approved):

AliceBilc (approved):

Ein-Tim:
@hilmarf 

Thanks for the information. 

First of all, this change is already live, is this intended? 

And second, do the information text shown in the app (tap on the "i" in the top right corner of a statistics card) also change according to the row of the statistic cards?

hilmarf:
Hi @Ein-Tim,

yes it's intended as we had last Thursday (July, 22nd) the production deployment of the pre-release https://github.com/corona-warn-app/cwa-server/releases/tag/v2.6.0
We normally first deploy newer backend version and have then the ability with very limited set of users to do crowd-testing against it. If everything is fine and no major issues were found and we had a release agreement with the RKI, then the staged rollout can begin for the mobile clients.

To your second point... actually I don't know. But I heard from the mobiles team, that the text/info screen behind the "i" is always the same and should not be affected by the order of the cards. It's more a static text page/legend.

Best regards,
Hilmar

Ein-Tim:
Thank you for your answers, @hilmarf!

PR #1515, opened 2021-07-30
encrypted checkins - Apple requirement

hilmarf:


common/persistence/src/main/java/app/coronawarn/server/common/persistence/service/TraceTimeIntervalWarningService.java
```diff
@@ -85,16 +91,49 @@ private int saveCheckins(List<CheckIn> checkins, Function<ByteString, byte[]> id
...
+          .saveDoNothingOnConflict(checkInProtectedReports.getLocationIdHash().toByteArray(),
+              checkInProtectedReports.getIv().toByteArray(),
+              checkInProtectedReports.getEncryptedCheckInRecord().toByteArray(),
+              submissionTimestamp);
+
+      if (checkInProtectedInsertedSuccessfully) {
+        numberOfCheckInProtectedReports++;
+      }
+    }
+    return numberOfCheckInProtectedReports;
```
hilmarf, line 113: should we raise an additional log message (WARN?) when the list.size() != numberOfCheckInProtectedReports

common/persistence/src/main/resources/db/specific/postgresql/V21__createTableForCheckInProtectedReports.sql
```diff
@@ -0,0 +1,13 @@
...
+    initialization_vector BYTEA NOT NULL,
+    encrypted_check_in_record BYTEA NOT NULL,
+    submission_timestamp BIGINT NOT NULL
+);
+
+GRANT INSERT ON TABLE check_in_protected_reports TO "cwa_submission";
+GRANT USAGE, SELECT ON SEQUENCE check_in_protected_reports_id_seq TO "cwa_submission";
+
+GRANT SELECT, DELETE ON TABLE check_in_protected_reports TO "cwa_distribution";
+GRANT USAGE, SELECT ON SEQUENCE check_in_protected_reports_id_seq TO "cwa_distribution";
```
hilmarf, line 13: does the distribution really use the `check_in_protected_reports_id_seq` sequence?

FelixRottler: Not sure, I will check it

common/persistence/src/main/java/app/coronawarn/server/common/persistence/service/TraceTimeIntervalWarningService.java
FelixRottler: I agree, but maybe only if number of unencrypted is greater than zero? @AliceBilc can you add this?

common/persistence/src/main/resources/db/specific/postgresql/V21__createTableForCheckInProtectedReports.sql
FelixRottler: Based on SELECT and DELETE it should not need the permissions right?

services/distribution/src/main/java/app/coronawarn/server/services/distribution/config/DistributionServiceConfig.java
```diff
@@ -227,6 +228,14 @@ public void setAndroidQrCodePosterTemplate(QrCodePosterTemplate androidQrCodePos
...
   }
 
+  public boolean isDeprecated() {
+    return isDeprecated;
+  }
+
+  public void setDeprecated(boolean deprecated) {
+    isDeprecated = deprecated;
+  }
+
```
hilmarf, line 238: what's this for?

services/distribution/src/test/java/app/coronawarn/server/services/distribution/assembly/tracewarnings/structure/directory/file/TraceTimeIntervalWarningExportFileTest.java
```diff
@@ -61,6 +61,7 @@ void fromTraceTimeIntervalWarningsShouldReturnNewInstance() throws Exception {
     byte[] checksum = Files.readAllBytes(outputFile.toPath().resolve(FILE_NAME_CHECKSUM));
     byte[] checksum2 = Files.readAllBytes(outputFile2.toPath().resolve(FILE_NAME_CHECKSUM));
 
+    //then
```
hilmarf, line 64: ?

FelixRottler: I see an assert statement after this comment. So I would close this comment

services/distribution/src/main/resources/application.yaml
```diff
@@ -113,8 +113,8 @@ services:
       bucket: ${STATISTICS_FILE_S3_BUCKET:obs-cwa-public-dev}
 
     app-features:
-      - label: reserved
-        value: ${reserved:1}
+      - label: unencrypted-checkins-enabled
+        value: ${EVREG_UNENCRYPTED_CHECKINS_ENABLED:0}
```
roesslerj, line 116: I think boolean (true/false) is easier to understand than 1/0 as is done in the submission yaml. Also decide for one.

roesslerj (dismissed):

services/distribution/src/main/resources/application.yaml
FelixRottler: @hilmarf  why do we have that? First time I see this this config and it is not used inside distribution.

hilmarf: it's the mobile app-feature toggle - becomes part of the mobile app-config, which is created by the distribution

FelixRottler: ok so that means there is nothing to do here right?

hilmarf: yes

FelixRottler (approved):

AliceBilc (approved):

PR #1526, opened 2021-08-10
ADD - audit logger for response body. dcc and dsc.

alexmoroianu:


services/distribution/src/main/java/app/coronawarn/server/services/distribution/dgc/client/ProdDigitalCovidCertificateClient.java
```diff
@@ -83,6 +85,9 @@ public ProdDigitalCovidCertificateClient(DigitalCovidCertificateFeignClient digi
     logger.debug("Get " + fetchEntityName + " from DCC");
     try {
       ResponseEntity<T> response = responseSupplier.get();
+
+      logger.info(fetchEntityName + "has been fetched and contains the following data: "
+          + stringifyObject(response.getBody()));
```
hilmarf, line 90: can this result again in an IOException (stream already closed) or something?

hilmarf:
@roesslerj - you know the log4j configs the best... is there any chance that we can specify another settings only for one or two special loggers?

@alexmoroianu - log4j2.xml limits the output to 100000 characters. Please double check, that the response payload 'normally' fits into that limit!

hilmarf (approved):

services/distribution/src/main/java/app/coronawarn/server/services/distribution/dgc/client/ProdDigitalCovidCertificateClient.java
alexmoroianu: yes, it can. I mean,  [Response execute(Request request, Options options)] only throws IOException. Even if the signature fails or any other reason. But I don't think it's possible to throw stream already closed anymore.

roesslerj:
@hilmarf: Yes, that's definitely possible. I will provide an example and get back to @alexmoroianu about that.

hilmarf (commented):
👍

roesslerj (approved):

PR #1533, opened 2021-08-17
Fix the wrong index file that is not recognised by CDN

roesslerj:
This is should already work... thanks to Felix ;-)

### DIFF
--- a/services/distribution/src/test/resources/application.yaml
+++ b/services/distribution/src/test/resources/application.yaml
@@ -8,7 +8,7 @@ logging:
 services:
   distribution:
     output-file-name: index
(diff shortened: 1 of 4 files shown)


ioangut (approved):

services/distribution/src/test/resources/application.yaml
```diff
@@ -8,7 +8,7 @@ logging:
 services:
   distribution:
     output-file-name: index
-    output-file-name-v2: index-v2
+    output-file-name-v2: index
```
ioangut, line 11: I think we can just get rid of this v2 env variable and use the normal one

PR #1535, opened 2021-08-17
fix: use correct length attributes for encrypted check in validation

FelixRottler:
- uses the correct length according to the spec https://github.com/corona-warn-app/cwa-app-tech-spec/pull/87/files#diff-b5b7febc8f64577750ae7bbe6cc6fc1aeffdec9d8b318556523861b1d82327e5R255


roesslerj (approved):
Can we have a test for that? Maybe create a ticket for creating a test?

AliceBilc (approved):

PR #1540, opened 2021-08-20
Added mac for encrypted checkins

AliceBilc:
## Checklist


common/persistence/src/main/resources/db/migration/V22_alterTableForCheckInProtectedReports.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE check_in_protected_reports ADD COLUMN mac BYTEA NOT NULL;
```
FelixRottler: This will be problem. All existing entries will have a null value but the constraint is not null.

FelixRottler (changes requested):
Also the length validation is missing for mac.
- the fix would be also a hotfix for 2.8

common/persistence/src/main/resources/db/migration/V22_alterTableForCheckInProtectedReports.sql
FelixRottler: Also I would rename this migration to reflect what changed about the table ( alterTableCheckInProtectedReportsAddMac or something like that

common/persistence/src/test/java/app/coronawarn/server/common/persistence/service/TraceTimeIntervalWarningServiceTest.java
```diff
@@ -181,9 +182,9 @@ void testSaveCheckInProtectedReports() {
   void testGetCheckInProtectedReports() {
```
FelixRottler, line 182: Feel like this test is at the wrong place. It is a test tracetimeintervalwarningSERVICE but you are testing the repository. So I would suggest adding a test for SERVICE layer there

AliceBilc: It seems to me that testSaveCheckInProtectedReports -> due to the fact that it checks the increased integer, it actually testing the service. I may be wrong, let's talk about it

FelixRottler: okay I agree, but then let's rename the test. Something like shouldReturnCheckInWithCorrectSubmissionOrder

mlenkeit:
Please keep in mind that this fix needs to go into 2.8

FelixRottler:
obsolete by other PR already merge on release/2.8

PR #1554, opened 2021-09-07
Feature/fix dcc it

roesslerj:
Missing UTF-* support is the reason this test sometimes failed ...

Given away by this String (when failing):
"desc": "Ne contient qu'un seul ??v??nement."
vs. this String (when passing):
"desc": "Ne contient qu'un seul évènement."


alexmoroianu (approved):

common/shared/src/main/java/app/coronawarn/server/common/shared/util/SecurityUtils.java
```diff
@@ -49,7 +49,11 @@ public static void ecdsaSignatureVerification(final byte[] encodedSignature, fin
     signatureVerification.update(content);
 
     if (!signatureVerification.verify(encodedSignature)) {
-      throw new SignatureException("Signature verification failed");
+      throw new SignatureException("Signature verification " + signatureVerification
```
FelixRottler, line 52: Maybe we want to use the algorithm instead of the object itself?

FelixRottler (changes requested):
When throwing signature exception, do we really want to include signatureVerification or is signatureVerification.getAlgorithm enough?

common/shared/src/main/java/app/coronawarn/server/common/shared/util/SecurityUtils.java
roesslerj: Good idea. Done :-)

FelixRottler (approved):

PR #1560, opened 2021-09-14
Booster notifications builder

alexmoroianu:


services/distribution/src/main/java/app/coronawarn/server/services/distribution/assembly/component/DigitalCertificatesStructureProvider.java
```diff
@@ -38,6 +45,7 @@
   public static final String DSCS = "dscs";
```
FelixRottler, line 45: I know it is not part of what you changed, but can we refactor also that name? What does it even mean?

PR #1574, opened 2021-10-04
Filter out unsupported card ids

ioangut:
## Checklist

## Description

Filter out unsupported card ids


services/distribution/src/main/java/app/coronawarn/server/services/distribution/statistics/StatisticsToProtobufMapping.java
```diff
@@ -155,18 +157,20 @@ private LocalDate effectiveDateStringToLocalDate(String effectiveDate) {
...
-            figureCardMap.put(id, Optional.of(card));
-          } catch (MissingPropertyException ex) {
-            logger.warn("[{}] {}", stat.getEffectiveDate(), ex.getMessage());
-          }
-        }
-      });
+      Arrays.stream(Cards.values()).map(card -> card.ordinal())
+          .filter(e -> getAllCardIdSequence().contains(e.intValue()))
+          .forEach(id -> {
+            if (figureCardMap.get(id).isEmpty()) {
```
mlenkeit, line 163: I think this does not reliably avoid NPEs for `figureCardMap.get(id).isEmpty()`. Technically, after the filter statement, there could still be a card id for which we forgot to put something into `figureCardMap`.

How about something like this:

```java
getAllCardIdSequence()
  .filter(id -> figureCardMap.has(id))
  .forEach(/*...*/)
```

We could also add a `warn` log in the filter statement if something is filtered out (or even error)

roesslerj (approved):

AliceBilc (approved):

PR #1609, opened 2021-11-18
Feature/remove unused code

roesslerj:
During working with the certificates for EFGS, I found some code that is obsolete and misleading. So rm.


roesslerj:
These parameters are actually used in some annotations.

PR #1630, opened 2021-12-09
Feature/app feature changes

roesslerj:
Contains two changes to the application.yaml as detailed here:
- Part 1: https://github.com/corona-warn-app/cwa-app-tech-spec/pull/145/files#diff-cb2505a644be1db6e9996aeca756a39fc510bdd5119eff62cc0e3d6a79046d02

- Part 2: https://github.com/corona-warn-app/cwa-app-tech-spec/pull/144/files#diff-cb2505a644be1db6e9996aeca756a39fc510bdd5119eff62cc0e3d6a79046d02

Has two separate commits...

### DIFF
--- a/services/distribution/src/main/resources/application.yaml
+++ b/services/distribution/src/main/resources/application.yaml
@@ -116,7 +116,7 @@ services:
       - label: unencrypted-checkins-enabled
         value: ${EVREG_UNENCRYPTED_CHECKINS_ENABLED:0}
       - label: validation-service-android-min-version-code
-        value: { $VALIDATION_SERVICE_ANDROID_MIN_VERSION_CODE:0 }
+        value: { $VALIDATION_SERVICE_ANDROID_MIN_VERSION_CODE:0}
       - label: validation-service-ios-min-version-major
         value: ${VALIDATION_SERVICE_IOS_MIN_VERSION_MAJOR:0}
       - label: validation-service-ios-min-version-minor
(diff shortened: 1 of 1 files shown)


AliceBilc (approved):

services/distribution/src/main/resources/application.yaml
```diff
@@ -115,6 +115,18 @@ services:
...
+      - label: validation-service-ios-min-version-major
+        value: ${VALIDATION_SERVICE_IOS_MIN_VERSION_MAJOR:0}
+      - label: validation-service-ios-min-version-minor
+        value: ${VALIDATION_SERVICE_IOS_MIN_VERSION_MINOR:0}
+      - label: validation-service-ios-min-version-patch
+        value: ${VALIDATION_SERVICE_IOS_MIN_VERSION_PATCH:0}
+      - label: dcc-person-count-max
+        value: ${DCC_PERSON_COUNT_MAX:20}
+      - label: dcc-person-warn-threshold
+        value: ${DCC_PERSON_WARN_THRESHOLD:10}
```
Ein-Tim, line 129: I'm getting the warning as soon as I'm trying to scan the certificates from more than 3 persons in the app. Has the threshold been lowered to 3 or is this a potential bug in the iOS implementation?

cc @dsarkar, If you could confirm that the threshold was lowered/not lowered, I will dig deeper and open an issue in the iOS repository.

dsarkar: @Ein-Tim I will check.

PR #1703, opened 2022-01-25
let's query only relevant TEKs from DB to lower memory footprint

hilmarf:
## Checklist

* [ ] Make sure `mvn -P integration-test clean verify` runs for the whole project and, if you touched any code for a service in the `services` folder, ensure it can be run with `spring-boot:run`

### DIFF
--- a/common/persistence/src/main/java/app/coronawarn/server/common/persistence/repository/DiagnosisKeyRepository.java
+++ b/common/persistence/src/main/java/app/coronawarn/server/common/persistence/repository/DiagnosisKeyRepository.java
@@ -1,8 +1,7 @@
-
-
 package app.coronawarn.server.common.persistence.repository;
 
 import app.coronawarn.server.common.persistence.domain.DiagnosisKey;
+import java.util.List;
 import org.springframework.data.jdbc.repository.query.Modifying;
 import org.springframework.data.jdbc.repository.query.Query;
 import org.springframework.data.repository.PagingAndSortingRepository;
@@ -80,4 +79,13 @@ public interface DiagnosisKeyRepository extends PagingAndSortingRepository<Diagn
       @Param("days_since_onset_of_symptoms") int daysSinceOnsetOfSymptoms,
       @Param("consent_to_federation") boolean consentToFederation,
       @Param("submission_type") String submissionType);
+
+  /**
+   * <code>SELECT * FROM diagnosis_key WHERE transmission_risk_level>=:minTRL SORT BY submission_timestamp</code>.
+   * 
+   * @param minTrl minimum Transmission-Risk-Level to be fetched from the database 
+   * @return List of {@link DiagnosisKey}s with given TRL or higher 
+   */
+  @Query("SELECT * FROM diagnosis_key WHERE transmission_risk_level>=:minTRL SORT BY submission_timestamp")
+  List<DiagnosisKey> findAllWithTrlGreaterThan(final @Param("minTrl") int minTrl);
 }
--- a/common/persistence/src/main/java/app/coronawarn/server/common/persistence/repository/DiagnosisKeyRepository.java
+++ b/common/persistence/src/main/java/app/coronawarn/server/common/persistence/repository/DiagnosisKeyRepository.java
@@ -81,11 +81,11 @@ public interface DiagnosisKeyRepository extends PagingAndSortingRepository<Diagn
       @Param("submission_type") String submissionType);
 
   /**
-   * <code>SELECT * FROM diagnosis_key WHERE transmission_risk_level>=:minTRL SORT BY submission_timestamp</code>.
-   * 
-   * @param minTrl minimum Transmission-Risk-Level to be fetched from the database 
-   * @return List of {@link DiagnosisKey}s with given TRL or higher 
+   * <code>SELECT * FROM diagnosis_key WHERE transmission_risk_level>=:minTRL ORDER BY submission_timestamp</code>.
+   *
+   * @param minTrl minimum Transmission-Risk-Level to be fetched from the database
+   * @return List of {@link DiagnosisKey}s with given TRL or higher
    */
-  @Query("SELECT * FROM diagnosis_key WHERE transmission_risk_level>=:minTRL SORT BY submission_timestamp")
+  @Query("SELECT * FROM diagnosis_key WHERE transmission_risk_level>=:minTrl ORDER BY submission_timestamp")
   List<DiagnosisKey> findAllWithTrlGreaterThanOrEqual(final @Param("minTrl") int minTrl);
 }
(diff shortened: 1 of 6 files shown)


common/persistence/src/main/java/app/coronawarn/server/common/persistence/repository/DiagnosisKeyRepository.java
```diff
@@ -80,4 +79,13 @@ boolean saveDoNothingOnConflict(
...
       @Param("submission_type") String submissionType);
+
+  /**
+   * <code>SELECT * FROM diagnosis_key WHERE transmission_risk_level>=:minTRL SORT BY submission_timestamp</code>.
+   * 
+   * @param minTrl minimum Transmission-Risk-Level to be fetched from the database 
+   * @return List of {@link DiagnosisKey}s with given TRL or higher 
+   */
+  @Query("SELECT * FROM diagnosis_key WHERE transmission_risk_level>=:minTRL SORT BY submission_timestamp")
+  List<DiagnosisKey> findAllWithTrlGreaterThan(final @Param("minTrl") int minTrl);
```
vaubaehn, line 90: hi @hilmarf ,
according to function name `findAllWithTrlGreaterThan`, the list will only contain dk > minTRL, but with regard to comments and the `select`, it's intended to get a list with dk >= minTRL. Could be good to align function name to prevent later confusion.

hilmarf: you are right... was bad naming

vaubaehn: @hilmarf 
Thanks :)

By the way, there seem to be some issues client side (probably Android only: repeated downloads). They should have been brought to CWA tech leads' attention already. If there's a chance to fix them, that could also reduce the memory footprint of transmissions to clients. Let's see...

vaubaehn: > By the way, there seem to be some issues client side (probably Android only: repeated downloads). They should have been brought to CWA tech leads' attention already. If there's a chance to fix them, that could also reduce the ~memory~ bandwidth footprint of transmissions to clients. Let's see...

@hilmarf 
Just to give you a more concrete idea what I was talking about, let's jump into the issue:
https://github.com/corona-warn-app/cwa-app-android/issues/4753#issuecomment-1025972121

hilmarf: @vaubaehn thanks for your inputs.

This current PR is more about an improvement of the jvm memory usage... it has not really something to do with reducing the load for the clients. That has been addressed already with previous PRs, where we quickly filtered the TEKs with TRL 1 and 2, which are anyhow irrelevant for the risk calculation on the clients. Unfortunately that quick fix is not very performant and can be done better, which is addressed by this PR, to optimize the operation runtime.

The architects are aware of your referenced ticket and are discussing further options to mitigate it... but that's not that easy... 

I hope that clarifies. 

Regards
Hilmar

vaubaehn: @hilmarf thanks for your kind reply. Things are clarified, and it's good to know for the community. that possible mitigations are under evaluation. Have a relaxing evening and stay safe! v.

ioangut (approved):

common/persistence/src/main/java/app/coronawarn/server/common/persistence/repository/DiagnosisKeyRepository.java
vaubaehn: Hi @hilmarf ,
As the situation is a bit unfortunate, I'm still investigating options on how to mitigate the issue of repeated downloads of diagnosis keys. So, here's my question:

Is there any possibility to query in advance the hash of a diagnosis key package provided by the server before downloading?
If yes, CWA could implement a logging for these hashes, and only download the packages that not have been downloaded and submitted to ENS before.
Could that be a way to solve this?
Kind regards, v.

mlenkeit: @vaubaehn there are ways to download files only if they are different from what the client downloaded last time (i.e. e-tag, if-none-match header, etc.). CWA is using this already in many places but for historical reasons is not using this for the key download. However, CWA is using other mechanisms that ensure that files are not repeatedly downloaded. We are stills struggling to understand why these repeated downloads happen on some devices and I'm not yet convinced that changing the caching approach in CWA will fix it.

Anyway, this PR here was not motivated by the repeated downloads but rather by the size of recent diagnosis key packages in general.

vaubaehn: @mlenkeit 
> @vaubaehn there are ways to download files only if they are different from what the client downloaded last time (i.e. e-tag, if-none-match header, etc.).

That was actually a way what I was thinking about and thus pointing my question to @hilmarf . Logging the hashes into a CWA-internal db as a control mechanism could then prevent superflous downloads of keys as well as non-necessary key provisions to ENF (as they're obviously not matched anymore by ENF due to a kind of suspected logging mechanism - but the inspection/file operations on provided key files need much time on older phones like mine thus contributing to timeouts in the workflow).

> However, CWA is using other mechanisms that ensure that files are not repeatedly downloaded.

If you have time, in one sentence, could you elaborate?

> We are stills struggling to understand why these repeated downloads happen on some devices and I'm not yet convinced that changing the caching approach in CWA will fix it.

I can share my observations: on older phones with less internal storage, the OS is wiping all app caches frequently to free some space. Hence DiagnosisKey-files are deleted (was confirmed by Fynn Godau in an internal chat, you might confirm by performing your own simple experiments), and if CWA's mechanism to prevent repeated downloads is not functionally here, it may lead to the observed problem. What might contribute to the problem of low memory is, that there might be a bug in ENF (https://github.com/corona-warn-app/cwa-app-android/issues/4732), but that's a weak suspect currently. Other observations indicate that former DK-RPI matches ("risk contacts") are cached by ENF and always returned by getExposureWindows() during the 14/15-day validity period. So it was a win on performance to only provide DKs to ENF that not have been submitted before. To confirm this behaviour, it would need more effort to experiment, as it either needs a whitelisted account or a rooted OS to inject crafted RPIs with matching TEKs (DKs) into ENF.
If my theory applies to reality, this possibly also mitigates issues where risk detection stops on older phomes (due to CWA and ENF timeouts) when incidence rates are high (resp. larger DK packages).

> Anyway, this PR here was not motivated by the repeated downloads but rather by the size of recent diagnosis key packages in general.

Yes, I got this from the beginning, but I wanted to share and exchange my thoughts also with @hilmarf , as I assumed he may contribute to a solution with his professional expertise. I hope everyone is fine with that, but we may also shift further discussion to the related Android issue.

PR #1709, opened 2022-02-01
Update LocalFile.java

ioangut:
## Description

The ccl-config files should be published when there is a change. This is not happening in the old implementation because the s3key.matches() , will mark the config-v1 as key file, which is incorrect.

### DIFF
--- a/services/distribution/src/main/java/app/coronawarn/server/services/distribution/objectstore/publish/LocalFile.java
+++ b/services/distribution/src/main/java/app/coronawarn/server/services/distribution/objectstore/publish/LocalFile.java
@@ -80,7 +80,7 @@ public abstract class LocalFile {
    */
   public String getContentType() {
     if (isConfigFile() || isStatisticFile() || isKeyFile() || isQrPosterTemplate() || isVaccineValueSet()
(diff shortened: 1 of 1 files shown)


AliceBilc (approved):

services/distribution/src/main/java/app/coronawarn/server/services/distribution/objectstore/publish/LocalFile.java
```diff
@@ -109,6 +109,10 @@ private boolean isStatisticFile() {
     return s3Key.endsWith("stats") || s3Key.matches(".*local_stats_\\d");
   }
 
+  private boolean isCclConfigFile() {
+    return s3Key.endsWith("stats") || s3Key.matches(".*config-v\\d");
```
hilmarf, line 113: why `endsWith("stats")` ???

```suggestion
    return s3Key.matches(".*config-v\\d");
```

ioangut: it is a mistake , thanks !

hilmarf (approved):

PR #1711, opened 2022-02-03
introducing: days-to-publish

hilmarf:


services/distribution/src/main/resources/application.yaml
```diff
@@ -20,6 +20,8 @@ services:
     output-file-name-v2: index-v2
     # The number of days to retain diagnosis keys and trace time warnings for both database persistency layer and files stored on the object store.
     retention-days: 14
+    # The number of days for diagnosis keys (based upon submission timestamp) to publish on CDN.
+    days-to-publish: ${DAYS_TO_PUBLISH:10}
```
Ein-Tim, line 24: This effectively means CWA does not take the last 14 days into account but only the last 10, correct?

hilmarf: yes - but still under discussion and not clear if it will come or not... but I'd like to be prepared when things get urgent :-)

Ein-Tim: Okay, thanks @hilmarf!

AliceBilc (approved):

ioangut (approved):

services/distribution/src/main/resources/application.yaml
Ein-Tim: Is this change now live? If yes, several texts in the app need updating, as well as the website.

Ein-Tim:
@hilmarf Please, could you let us know if this change is live already?

Ein-Tim:
@hilmarf 

Ok, thanks! Just to be sure, on https://svc90.main.px.t-online.de/version/v1/diagnosis-keys/country/EUR/date I see the dates 2022-01-27 - 2022-02-06

So the last 11 days, without today. So are the last 11 days considered, or?

FYI @vaubaehn

hilmarf:
good catch!

yeah, the HF was deployed with wrong environment setting :-(

now it has been corrected and https://svc90.main.px.t-online.de/version/v1/diagnosis-keys/country/EUR/date shows only last 10 days

Ein-Tim:
@hilmarf 

Okay, nice, so only the last 10 days are taken into account now. Thank you!

vaubaehn:
> good catch!
> 
> yeah, the HF was deployed with wrong environment setting :-(
> 
> now it has been corrected and https://svc90.main.px.t-online.de/version/v1/diagnosis-keys/country/EUR/date shows only last 10 days

confirmed :wink Thanks for the quick fix.

PR #1736, opened 2022-02-21
Add dcc-admission-check-scenarios-disabled in appFeatures

AliceBilc:
## Checklist

### DIFF
--- a/services/distribution/src/main/resources/application.yaml
+++ b/services/distribution/src/main/resources/application.yaml
@@ -132,7 +132,7 @@ services:
       - label: dcc-person-warn-threshold
         value: ${DCC_PERSON_WARN_THRESHOLD:4}
       - label: dcc-admission-check-scenarios-disabled
-        value: ${DCC_ADMISSION_CHECK_SCENARIOS_ENABLED:0}
+        value: ${DCC_ADMISSION_CHECK_SCENARIOS_DISABLED:1}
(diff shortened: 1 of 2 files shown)


ioangut (approved):

services/distribution/src/main/resources/application.yaml
```diff
@@ -131,6 +131,8 @@ services:
         value: ${DCC_PERSON_COUNT_MAX:20}
       - label: dcc-person-warn-threshold
         value: ${DCC_PERSON_WARN_THRESHOLD:4}
+      - label: dcc-admission-check-scenarios-disabled
+        value: ${DCC_ADMISSION_CHECK_SCENARIOS_ENABLED:0}
```
hilmarf, line 135: @mlenkeit this will be the same for iOS and Android - is that OK? And `0` is the right default as of now, correct?

mlenkeit: @AliceBilc I would propose the following (two) changes: align the env var name with the `label` attribute and then set it to `1` by default.
```suggestion
      - label: dcc-admission-check-scenarios-disabled
        value: ${DCC_ADMISSION_CHECK_SCENARIOS_DISABLED:1}
```

mlenkeit (changes requested):
@AliceBilc see inline comment

mlenkeit (approved):
LGTM!

hilmarf (approved):

PR #1749, opened 2022-02-28
Add new app config parameters

AliceBilc:
## Checklist


AliceBilc:
@mlenkeit, we're already using this parameter in the filter method https://github.com/corona-warn-app/cwa-server/blob/main/services/submission/src/main/resources/application.yaml#L57 
Do you think it would be a good ideea to change here the name and the value?

hilmarf (approved):

ioangut (approved):

PR #1809, opened 2022-04-26
fix NPE issue with http HEAD

hilmarf:
when ETag is fetched, feign is failing with NullPointerException, because the response body is empty/null :-(

also moved the DB truncation away from 'retention' into the 'assembly' to have one single transaction - in case of failures, we'll be at least consistent then ;-)


services/distribution/src/main/java/app/coronawarn/server/services/distribution/dcc/ProdDccRevocationClient.java
```diff
@@ -64,8 +65,8 @@ public String getETag() throws FetchDccListException {
    * @throws NullPointerException if there is no ETag in the {@link ResponseEntity#getHeaders()}.
    */
   public static String getETag(final ResponseEntity<?> response) {
-    final String string = response.getHeaders().getETag().replaceAll("\"", "");
```
AliceBilc, line 67: Why was the replaceAll("\"", "") deleted?

AliceBilc (approved):

PR #1821, opened 2022-05-03
switch from s3 ListObjectsV2 to ListObjects

hilmarf:
Object Store doesn't support V2 API with token... :-(

When DCC Revocation is running, we need to query ALL objects from OBS buckets and delete the same, incase the ETag did change and we need to republish everything.

But in case of regular 'distribution' run, we should ignore all S3 objects with the delimiter `/dcc-rl/`, to speed up things.

### DIFF
--- a/services/distribution/src/main/java/app/coronawarn/server/services/distribution/objectstore/client/S3ClientWrapper.java
+++ b/services/distribution/src/main/java/app/coronawarn/server/services/distribution/objectstore/client/S3ClientWrapper.java
@@ -2,6 +2,7 @@ package app.coronawarn.server.services.distribution.objectstore.client;
 
 import static java.lang.Boolean.TRUE;
 import static java.util.stream.Collectors.toList;
+import static software.amazon.awssdk.services.s3.model.ListObjectsRequest.builder;
 
 import app.coronawarn.server.services.distribution.statistics.exceptions.NotModifiedException;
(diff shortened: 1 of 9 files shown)


ioangut (dismissed):

services/distribution/src/main/java/app/coronawarn/server/services/distribution/objectstore/client/S3ClientWrapper.java
```diff
@@ -47,8 +49,16 @@ public class S3ClientWrapper implements ObjectStoreClient {
...
   private final S3Client s3Client;
 
+  private final String dccListPath;
+
   public S3ClientWrapper(S3Client s3Client) {
     this.s3Client = s3Client;
+    this.dccListPath = null;
+  }
+
+  public S3ClientWrapper(final S3Client s3Client, final String dccListPath) {
```
mlenkeit, line 59: @hilmarf on this level of abstraction, shouldn't this rather be exposed as `delimiter` instead of `dccListPath`? I'm sure future me will wonder why we need `dccListPath` here on this level. Btw, I would suggest to rename this to something that reflects that it's about _DCC revocation_

hilmarf: added now more JavaDoc

ioangut (approved):

PR #1930, opened 2022-10-10
merge pandamic radar from main into Release 2.28

hilmarf:
#1926

### DIFF
--- a/services/distribution/src/main/resources/application.yaml
+++ b/services/distribution/src/main/resources/application.yaml
@@ -38,7 +38,7 @@ services:
     eu-package-name: EUR
     # Indicates whether the shifting and expiry policies are applied to all supported countries during distribution.
     apply-policies-for-all-countries: false
-    card-id-sequence: ${STATS_CARD_ID_SEQUENCE:[10,2,8,9,1,3,4,5,6,11,7]}
+    card-id-sequence: ${STATS_CARD_ID_SEQUENCE:[12,999,10,2,8,9,1,3,4,5,6,11,7]}
     # Local paths, that are used during the export creation.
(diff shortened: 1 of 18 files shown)


mfromme (approved):

services/distribution/src/main/resources/application.yaml
```diff
@@ -38,7 +38,7 @@ services:
     eu-package-name: EUR
     # Indicates whether the shifting and expiry policies are applied to all supported countries during distribution.
     apply-policies-for-all-countries: false
-    card-id-sequence: ${STATS_CARD_ID_SEQUENCE:[10,2,8,9,1,3,4,5,6,11,7]}
+    card-id-sequence: ${STATS_CARD_ID_SEQUENCE:[12,999,10,2,8,9,1,3,4,5,6,11,7]}
```
Ein-Tim, line 41: Why is "999" in here?

Ein-Tim (commented):
One question

services/distribution/src/main/resources/application.yaml
hilmarf: @Ein-Tim - good catch :-)

It's a kind of 'spacer' - to ensure the new card with the link to the pandemic radar, will be always the first one, regardless how many 'local' static tiles a user has configured. Normally the mobile app would place the stats of the region/local at first.

@mlenkeit - f.y.i. in case you want to add...

hilmarf: > Also, was the decision to show the "Pandemieradar" at first place, made based on customer requirements or is this something which the team decided?

It's customer requirement.

PR #1949, opened 2022-11-01
Self-Report Submission (SRS)

hilmarf:
https://github.com/corona-warn-app/cwa-ppa-server/pull/564


common/protocols/src/main/proto/app/coronawarn/server/common/protocols/internal/submission_payload.proto
```diff
@@ -19,5 +19,11 @@ message SubmissionPayload {
     SUBMISSION_TYPE_PCR_TEST = 0;
     SUBMISSION_TYPE_RAPID_TEST = 1;
     SUBMISSION_TYPE_HOST_WARNING = 2;
+    SUBMISSION_TYPE_SRS_SELF_TEST = 3;
+    SUBMISSION_TYPE_SRS_RAT = 4;
+    SUBMISSION_TYPE_SRS_REGISTERED_PCR = 5;
+    SUBMISSION_TYPE_SRS_UNREGISTERED_PCR = 6;
+    SUBMISSION_TYPE_SRS_RAPID_PCR = 7;
+    SUBMISSION_TYPE_SRS_OTHER = 8;
```
Ein-Tim, line 27: If this feature is what I think it is (warning without a QR code / a TAN), I find it quite strange that there are two categories `REGISTERED_PCR` & `UNREGISTERED_PCR`
but only `RAT` and not `REGISTERED_RAT` & `UNREGISTERED_RAT`. Why is that?

thomasaugsten: This is only a proof of concept. There is no official feature planned at the moment.
At the moment we think all registered RAT test should be warnable via qr code. But as  I mentioned this is at the moment only ideation

Ein-Tim: Maybe you should still consider making two options also for RATs, this would give further insights how reliable testing stations deliver test results. Should this feature actually be delivered one day (-;

Ein-Tim: @thomasaugsten 

Here for example a user on Twitter who registered a RAT but the test result never actually arrived in the CWA: https://twitter.com/lohbi/status/1589856359817785345?s=46&t=0wrqSZF_f46E38TKx-bTNA

Also GitHub issue: https://github.com/corona-warn-app/cwa-testresult-server/issues/138 (Title is wrong, it's about RATs).

I don't really see the point of only going half way here and not go all in, to be honest. More options don't do any harm and give more data insight and may reveal hidden problems.

thomasaugsten: We will add registered_rat to the list

mlenkeit: see https://github.com/corona-warn-app/cwa-server/pull/1955

Ein-Tim: Thank you @thomasaugsten & @mlenkeit!

services/submission/src/main/java/app/coronawarn/server/services/submission/controller/SubmissionController.java
```diff
@@ -88,16 +97,75 @@
...
+    if (!isEmpty(tan) && !isEmpty(otp)) {
+      logger.warn("'cwa-authorization' and 'cwa-otp' header provided!");
+      return badRequest();
+    }
+
+    if (!isEmpty(otp)) {
+      if (!validSrsType(exposureKeys.getSubmissionType().getNumber())) {
+        return badRequest();
+      }
+      logger.debug("SRS: {} - {}", otp, exposureKeys);
```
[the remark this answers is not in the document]
hilmarf: https://github.com/corona-warn-app/cwa-server/pull/1949/commits/aa01812aa45095cb1f5d5a62e98bb2039787a59a

mfromme (approved):

Ein-Tim:
Is the value for how long the app needs to be installed before a SRS warning can be issued anywhere available in the code or will it stay secret for security reasons?

hilmarf:
> cret for security rea

Hi @Ein-Tim,

as far as I know the time how long an app needs to be installed isn't really a hard requirement. At least the backend doesn't know it and can't check it.

But we've implemented a limitation at [ppac](https://github.com/corona-warn-app/cwa-ppa-server/blob/feature/srs/services/ppac/src/main/resources/application.yaml#L69) `SRS_TIME_BETWEEN_SUBMISSIONS_IN_DAYS`, which is currently defined by RKI to be 90 days. Means: Android/iOS devices are checked against device specific validations (to ensure, they aren't rooted etc.) and can then save a One-Time-Passcode (OTP), which can be used in a second backend-call to the `submission` service to upload the TEKs. This OTP creation can happen only every 90 days (at the moment). Depending on RKI analysis and evolution of the pandemic, we might need adjust the value.

I hope this answers your question.
Happy weekend,
Hilmar

Ein-Tim:
Hey @hilmarf,

Thanks for the answer! Yes, this answers my question! If I'm not mistaken, the frontend apps have implemented a threshold check for how long the CWA has been installed until the user can warn someone using a SRS. But I will try to find more information on this myself.

> But we've implemented a limitation at [ppac](https://github.com/corona-warn-app/cwa-ppa-server/blob/feature/srs/services/ppac/src/main/resources/application.yaml?rgh-link-date=2022-12-09T16%3A11%3A06Z#L69) `SRS_TIME_BETWEEN_SUBMISSIONS_IN_DAYS`, which is currently defined by RKI to be 90 days. Means: Android/iOS devices are checked against device specific validations (to ensure, they aren't rooted etc.) and can then save a One-Time-Passcode (OTP), which can be used in a second backend-call to the `submission` service to upload the TEKs. This OTP creation can happen only every 90 days (at the moment). Depending on RKI analysis and evolution of the pandemic, we might need adjust the value.

Ah, interesting! Thanks for sharing details and explaining this security measure!

Have a nice weekend, and thanks for your great work!
Tim

mfromme (approved):

